PR #23, opened 2019-03-18
Update readme to use 'npm install'

pentacular:


BarbourSmith:
I'm getting the error message:

image

When testing the npm install method 😢

pentacular:
Looks like esm was misspelt as exsm. :)

PR #24, opened 2019-03-18
Begin quick start guide

BarbourSmith:
This pull request isn't complete, but it's a beginning of what I think would be useful as a quick start guide.

How can I find out more about how the website https://jsxcad.js.org/ uses three.js (?) to render the output from JSXCAD?

**Edit:** My understanding is that running the command writeThreejsPage() will generate a.html file which is self contained (is it?) and can be embedded in a web page using an iframe. If that's all there is to it...it's brilliant and very simple.

**Edit2:** It looks like it isn't fully self contained, but downloads the scripts it needs automatically which is very cool. If I regenerate the .html file and refresh the iframe the scrips will be cashed and not reload, right? It looks like there is a TODO about abstracting the file system so that it can work in a browser, is that a lot of work?


BarbourSmith:
In fact, would it be possible to have the https://jsxcad.js.org/ website exist in a repo somewhere? That way we could make pull requests to modify it, and also it would serve as an example.

I would be more than happy to spend a little time cleaning up the website (even though I know it's in a very basic form right now and will change completely)

pentacular:
https://jsxcad.js.org/ is https://github.com/jsxcad/jsxcad.github.io

writeThreejsPage() is a quick hack to test threejs integration.
It may be useful in displaying static models, but I don't expect people to be running writeThreejsPage in the browser normally.

What I'm imagining is something more like writeSolid({ path: 'myproject/foo' }, model)

And then if you're on a server, that might write to a file, but if you're on a browser you might have it effectively write to a viewer window, which would then update with the new model data.

That might look a bit like what writeThreejsPage does, except that instead of hard-coding the geometry data into the page, you'd just send it along dynamically.

And then you might imagine having a number of these windows to display different models, which might be parts of the same assembly or different assemblies, and might be laser cut pathguides, or surfaces, etc.

As for abstracting the file-system, I think it's mostly just a little plumbing work, so not very hard.

PR #35, opened 2019-03-24
cube: added corner1/corner2 parameters (openjscad compat).

fischman:
Also:
- Added an example+test utilizing the above API change.
- Unified handling of roundRadius & resolution in cube's params.

### DIFF
--- a/api/v1/cube.js
+++ b/api/v1/cube.js
@@ -26,7 +26,7 @@ const buildCube = ({ center = false, centerPosition, length = 1, width = 1, heig
   }
   return cube;
 };
-    
+
 const decode = (params) => {
   // cube(10)
   try {
@@ -48,10 +48,10 @@ const decode = (params) => {
     Object.assign(roundParams, { round: true, roundingRadius: roundRadius, roundingFaces: resolution });
   }
 
-  const decoded = (() => {
+  const decodeVariants = () => {
     // cube({ center: [0, 0, 0], radius: 1 })
     try {
-      const { center } = params[0];
+      const { center, radius } = params[0];
       assertNumberTriple(center);
       assertNumber(radius);
       assertSingle(params);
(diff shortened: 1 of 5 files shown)


api/v1/cube.js
```diff
@@ -96,7 +42,79 @@ const decode = (params) => {
...
   } catch (e) {}
 
-  throw Error(`Unsupported interface for cube: ${JSON.stringify(params)}`);
+  const { roundRadius = undefined, resolution = 8 } = params[0];
+  const roundParams = { round: false };
+  if (typeof roundRadius === 'number' && typeof resolution === 'number') {
+    Object.assign(roundParams, { round: true, roundingRadius: roundRadius, roundingFaces: resolution });
+  }
+
+  const decoded = (() => {
```
pentacular: I thought this was defining a function until I read all the way to the end. :)

Can we separate the function definition from its application?

pentacular:
Thanks. 

Can you
    npm run lint
at the top level and then run
    eslint --fix *.js
in the directories where lint is unhappy?

When you fix that the travis checks should pass. :)

fischman:
De-linted, hopefully.
eslint isn't a command my shell knows, and googling gave me this solution:
../../node_modules/.bin/eslint --fix *.js
(only mentioning in case you want to add to docs)

pentacular:
Yeah, I think we need per package support for 'npm test' and 'npm run lint'.

I've mostly skipped over it because I installed these tools globally a while back, and so it hasn't affected me -- but it's not a good situation, and prone to version skew and so on.

example/v1/interlock-proof-of-concept.js
```diff
@@ -0,0 +1,15 @@
...
+    cube({ corner1: [0, 1, 1], corner2: [40, 3, 2.5] }),
+
+    cube({ corner1: [0, 10, 0], corner2: [40, 14, 1] }),
+    cube({ corner1: [0, 10, 1], corner2: [40, 11, 2.5] }),
+    cube({ corner1: [0, 13, 1], corner2: [40, 14, 2.5] })
+  ]);
+};
+writeThreejsPage({ path: '/tmp/interlock-proof-of-concept.html', cameraPosition: [0, 100, 100] }, main());
+
+export const writeStlForTest = () => writeStl({ path: 'tmp/interlock-proof-of-concept.stl' }, main());
```
pentacular: Yeah, I should fix up that jscad runner to do writes from the return of main automatically. :)

Sorry about that.

PR #49, opened 2019-03-29
Add instructions to run locally to readme

BarbourSmith:
This might seem obvious, but I've seen at least one other question about how to do this in the forums and giving people a really easy and clear path to start will encourage more development


pentacular:
Thanks. :)

PR #54, opened 2019-03-31
Switch writing to screen to a paths based approach part 1

BarbourSmith:
## What this pr does

This pull request restructures how writing to the screen through three.js works. It adds two new methods to the api which are `writeScreen` and `clearScreen`. 

`writeScreen({path : 'window/viewer/thing1'}, sphere(2).translate([0, 0, -1.5]));` will add a threejs object with tag 'thing1' to the threejs instance attached to the domElement with id `viewer`. If the element with id viewer doesn't have a threejs instance attached to it one will be created. The thinking is that the behavior should be similar to writePDF which will write to the file specified by the path. If the file doesn't already exist it will be created.

`clearScreen({path : 'window/viewer/thing1'});` will remove thing1 from the targeted viewer and `clearScreen({path : 'window/viewer/*'});` will remove everything from the targeted viewer.

## What doesn't this pr do

Currently two things are not working in this pr. 

We talked about the need for the ability to have multiple viewers on the screen at once, possibly in a "picture in picture" mode.  This gets us started in that direction because we can place as many threejs instances as we want easily, but right now the threejs controls are bound to the entire window so all the views move at once, ~~the sizing is also wrong~~(Edit: fixed). I will fix this behavior in a follow up pull request.

Assemblies render correctly in this pull request, but information is lost along the way about the materials of individual parts, and there is not a way right now to show and hide individual parts. Adding those two behaviors won't be particularly difficult, but I think we want to think a little bit about how we want it to be done.


BarbourSmith:
I feel like I am more likely to break something than to fix something with the lint system so I'm going to let it be for now and update from master before linting this pr

api/v1/writeScreen.js
```diff
@@ -0,0 +1,15 @@
+import { buildThreejsScreen } from '@jsxcad/algorithm-threejs';
+
+export const writeScreen = ({ path }, ...shapes) => {
```
pentacular: As this is part of the API, my first question will be -- what would this do when running in node? :)

Instead of this, I'd suggest writing a geometry to a pathname.

e.g., writePdf({ path: 'gear.stl' }, gear);

This will then work on node.

And then on the browser, using

import { watchFile } from '@jsxcad/sys';
watchFile('gear.stl', renderGeometry); // or something

Now renderGeometry should get called with a 'file' object when the writePdf happens.

The file object should contain 'data' which is the geometry, and a 'translator' which will produce the pdf file content if you call it.

So now, you can convert the 'data' for threejs and display it in the appropriate window for showing a path like 'gear.stl'.
(And call the translate function later on if the user presses a download button or whatever).

So far, this should already exist, but there are a couple of missing bits.

There should be a way to watch for new files.

Then you could say something like watchNewFiles(watcher); and have watcher call watchFile above to catch the data.

Which leaves the missing part of how to map files to windows.

This is something that probably makes sense to add to the api -- and this should be safe because it will have compatible behavior in the browser as well as in node.

I'm not sure what the best approach is, but I guess I'd suggest starting with something like a regular expression matcher.

e.g., ".*" to "main window" or "gear/.*" to "gear window".

Maybe something like

addPreview({ pathPattern: "gear/.*", windowName: "gear" })
addPreview({ pathPattern: "cuts/.*", windowName: "cuts" })

...

writePdf({ path: "cuts/profile.pdf", profile);
writeStl({ path: "gear/preview.stl", gear);

The main thing is that the same script should work both on server and in browser -- and I don't see how that's possible with writeScreen.

BarbourSmith: I agree, that would be very cool. It could also be file type agnostic so you could write a pdf and have it preview nicely in the browser or write an stl file and see a 3D shape. That's definitely getting beyond the things I have experience in so I might not be the right person to do it well. Also since my use case is strictly browser based I'm probably not the best person for the job because I don't really know what a node user would want or not want...but in theory I see what you are saying and I agree that would be the right approach. Obviously we need a solution which works in both node and the browser.

api/v1/clearScreen.js
```diff
@@ -0,0 +1,17 @@
...
+    //Clear screen if it exists at path
+    var threeInstance = document.getElementById(targetDiv).threeScreen;
+    
+    //Add a viewer to the path if there isn't one already
+    if(typeof threeInstance != 'undefined'){
+        if(shapeID == '*'){
+            threeInstance.clearScreenAll();
+        }
+        else{
+            threeInstance.clearScreenById(shapeID);
```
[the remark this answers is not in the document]
BarbourSmith: Good suggestion! I agree that is more clear

api/v1/clearScreen.js
```diff
@@ -0,0 +1,17 @@
+export const clearScreen = ({ path }) => {
```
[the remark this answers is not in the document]
BarbourSmith: I'm not sure what this would do when running in node 😁... probably crash 😥...what would be the desired behavior?

Writes are cumulative right now so you can keep writing shapes to the display without clearing so writing an empty shape would do just that. My thinking was that you might want to be able to recompute and redraw just part of a model without touching the rest, but maybe that's not the right behavior

pentacular:
I've updated app/v1 in the simplest way to support multiple windows by way of illustration.

PTAL, and see if it makes things a bit more obvious. :)

algorithm/threejs/buildThreejsScreen.js
```diff
@@ -0,0 +1,129 @@
...
+    // stats.update();
+  }
+
+  render () {
+    this.renderer.render(this.scene, this.camera);
+  }
+}
+
+export const buildThreejsScreen = (targetDiv) => {
+  new ThreejsDisplay(targetDiv);
```
pentacular: Probably need a return or to remove the {} :)

package.json
```diff
@@ -6,6 +6,7 @@
   "scripts": {
     "bootstrap": "lerna bootstrap --hoist",
     "lint": "lerna bootstrap --hoist && lerna run lint",
+    "lintfix": "lerna bootstrap --hoist && lerna run lintfix",
```
pentacular: Thanks for fixing this :)

algorithm/threejs/buildThreejsScreen.js
```diff
@@ -0,0 +1,129 @@
...
+    this.onWindowResize();
+    // gui.add( material, 'wireframe' );
+    //
+    window.addEventListener('resize', () => { this.onWindowResize(); }, false);
+
+    this.onWindowResize();
+    this.animate();
+  }
+
+  writeScreen (options = {}, ...shapes) {
```
pentacular: Ok, this makes sense here :)

algorithm/threejs/package.json
```diff
@@ -54,6 +54,7 @@
   ],
   "license": "MIT",
   "dependencies": {
+    "@jsxcad/algorithm-polygons": "^0.0.50",
```
pentacular: I guess you'll need to add a three dependency here?

package.json
BarbourSmith: Thanks for telling me how to 😀

BarbourSmith:
So after we talked yesterday I'm kind of thinking that this pull request doesn't seem like the right direction to go. I like your idea of using the same writeSTL() functions to write to a path and then just set up a watcher for the default path in the index.html file and handle the threejs things there. 

One of my goals with this pull request was to get as much javascript out of the index.html file as possible because that felt cleaner and it seemed like that would reduce the duplication of code when JSxCAD is used on multiple websites, but I think I was wrong about that.. It makes sense to me that the threejs data processing stuff in algorithm is there, but I think the index.html file is actually the right place for the rest of it. 

If it's OK with you I'd like to close this pr and treat it as a learning experience and cherry pick the parts I like from it for a new one to implement using writeSTL to write to the screen with a paths approach. It shouldn't take much work beyond what you have already done. 

I don't see how I would be able to implement the multiple pages approach that we both played with, but I'm ok with that if you are.

pentacular:
> So after we talked yesterday I'm kind of thinking that this pull request doesn't seem like the right direction to go. I like your idea of using the same writeSTL() functions to write to a path and then just set up a watcher for the default path in the index.html file and handle the threejs things there.
> 
> One of my goals with this pull request was to get as much javascript out of the index.html file as possible because that felt cleaner and it seemed like that would reduce the duplication of code when JSxCAD is used on multiple websites, but I think I was wrong about that.. It makes sense to me that the threejs data processing stuff in algorithm is there, but I think the index.html file is actually the right place for the rest of it.

Honestly, I have no idea :)
I just stuck it in there since it seemed like the easiest thing to do to get to the next bit.

I'm happy with leaving it there until we find a better place to put it. :)

> 
> If it's OK with you I'd like to close this pr and treat it as a learning experience and cherry pick the parts I like from it for a new one to implement using writeSTL to write to the screen with a paths approach. It shouldn't take much work beyond what you have already done.

Sure, although writeStl already works.

For some reason I forgot to change writePaths to writeStl in the example. :(

I guess what's missing is adding an implicit writeStl({ path: 'output.stl', ...) for the return of main.
(is output.stl a reasonable name?)
(And in the future, I guess we'll want to guess when Stl or Svg or whatever is the right choice depending on how flat the paths are)

And then the rest is the hard work of making a sensible UI that does not fill people with rage and confusion. :)

> 
> I don't see how I would be able to implement the multiple pages approach that we both played with, but I'm ok with that if you are.

BarbourSmith:
> Sure, although writeStl already works.
> 
> For some reason I forgot to change writePaths to writeStl in the example. :(

Woah, that is fantastic. Awesome. Well then my goal for the day is complete 😀 

I would still like to spend some time to better understand how it works so that's what I'm going to do now

> I guess what's missing is adding an implicit writeStl({ path: 'output.stl', ...) for the return of main.
(is output.stl a reasonable name?)
(And in the future, I guess we'll want to guess when Stl or Svg or whatever is the right choice depending on how flat the paths are)

Forgive me if I'm misunderstanding, but maybe we would want the default path to be something like defaultOutput without the .stl on the end because then we could write both stl and svg or anything else to that path and have it display...am I thinking about that correctly?  Although then when you run the same code in node you get a invalid file without an extension

Again, awesome work getting it working already. I just played around with it in the web view and it works exactly the way I would expect it to 👍  👍

pentacular:
On Wed, Apr 3, 2019, 2:14 PM BarbourSmith <notifications@github.com> wrote:

> Sure, although writeStl already works.
>
> For some reason I forgot to change writePaths to writeStl in the example.
> :(
>
> Woah, that is fantastic. Awesome. Well then my goal for the day is
> complete 😀
>
> I would still like to spend some time to better understand how it works so
> that's what I'm going to do now
>
Hopefully you'll find all of the bits I did wrong. :)

> I guess what's missing is adding an implicit writeStl({ path:
> 'output.stl', ...) for the return of main.
> (is output.stl a reasonable name?)
> (And in the future, I guess we'll want to guess when Stl or Svg or
> whatever is the right choice depending on how flat the paths are)
>
> Forgive me if I'm misunderstanding, but maybe we would want the default
> path to be something like defaultOutput without the .stl on the end because
> then we could write both stl and svg or anything else to that path and have
> it display...am I thinking about that correctly? Although then when you run
> the same code in node you get a invalid file without an extension
>
Ah, that's an interesting point.

At least in the node, we can supply command line arguments saying what the
default format and filename should be, which can default to something
reasonable.

e.g., run foo --format stl --output foo.stl

In the browser we can defer the decision about file-format until they
actually want to download it.

So maybe writePaths is good for something after all. :)

The main thing that complicates this is that some kinds of geometry don't
make sense for some kinds of output.
e.g., stl should have solids, svg, pdf, dst should have z0 paths.

The other complication is that there will be per-format options, but I
guess those can be handled by command line arguments or menus on download.

> Again, awesome work getting it working already. I just played around with
> it in the web view and it works exactly the way I would expect it to 👍 👍
>
Cool.

btw, I'm not sure that the page flipping approach is a great one -- it just
seemed like the shortest path to getting multiple displays working.

I'm going to work on cleaning up assemblies for a while -- they need to be
made fully lazy to avoid defective geometry.

At some point we should also add support for displaying flat surfaces (like
cross-sections) and paths. :)

BarbourSmith:
> The main thing that complicates this is that some kinds of geometry don't
make sense for some kinds of output.
e.g., stl should have solids, svg, pdf, dst should have z0 paths.

I was kind of thinking it would be really cool to be able to pass a solid to writePDF() and get back a line drawing pdf of the solid. For example if you wanted to embroider a 3D house you would model a 3D house, and then call writePDF to get a line drawing of it. Cad programs like SolidWorks usually have this type of feature for generating blueprints and drawings which show the context that things go together. It might be complicated to implement.

> At some point we should also add support for displaying flat surfaces (like
cross-sections) and paths. :)

I played around with this a little bit and it seems like the 2d shapes like circle() render correctly as a 2D plane. I wasn't able to extrude them to make a 3D shape however, but I may have just been doing it wrong.

> Hopefully you'll find all of the bits I did wrong. :)

I went through everything to understand it and then got a version working in my flow drawing program and everything looks great. The more I think about it the more I like the way you have done things with the paths and watching approach. That makes it very easy for me to write to the screen from anywhere in my program without having to pass around a pointer or global object or anything like that. Well done!

For the sake of making some suggestion (albeit not a major one), the only thing I had any reservations about is having the materials library defined in makeMaterial(). That seems to me like something which is going to need to be duplicated with each implementation. I'm thinking either the material property of the shape contains information about it's color and roughness..etc, or we have a dictionary which is exposed through the API that keeps track of what 'metal' means so we can look it up in the .html file.

pentacular:
On Wed, Apr 3, 2019 at 3:26 PM BarbourSmith <notifications@github.com>
wrote:

> The main thing that complicates this is that some kinds of geometry don't
> make sense for some kinds of output.
> e.g., stl should have solids, svg, pdf, dst should have z0 paths.
>
> I was kind of thinking it would be really cool to be able to pass a solid
> to writePDF() and get back a line drawing pdf of the solid. For example if
> you wanted to embroider a 3D house you would model a 3D house, and then
> call writePDF to get a line drawing of it. Cad programs like SolidWorks
> usually have this type of feature for generating blueprints and drawings
> which show the context that things go together. It might be complicated to
> implement.
>
> At some point we should also add support for displaying flat surfaces (like
> cross-sections) and paths. :)
>
> I played around with this a little bit and it seems like the 2d shapes
> like circle() render correctly as a 2D plane. I wasn't able to extrude them
> to make a 3D shape however, but I may have just been doing it wrong.
>
Best I can get is writeStl({ path: 'assembly' }, circle()); which looks
like it's trying to turn the circle arcs into triangles. :)

but writeStl({ path: 'assembly' }, circle().extrude({ height: 1 })) should
work for extrusion.

> Hopefully you'll find all of the bits I did wrong. :)
>
> I went through everything to understand it and then got a version working
> in my flow drawing program and everything looks great. The more I think
> about it the more I like the way you have done things with the paths and
> watching approach. That makes it very easy for me to write to the screen
> from anywhere in my program without having to pass around a pointer or
> global object or anything like that. Well done!
>
> For the sake of making some suggestion (albeit not a major one), the only
> thing I had any reservations about is having the materials library defined
> in makeMaterial(). That seems to me like something which is going to need
> to be duplicated with each implementation. I'm thinking either the material
> property of the shape contains information about it's color and
> roughness..etc, or we have a dictionary which is exposed through the API
> that keeps track of what 'metal' means so we can look it up in the .html
> file.
>
Yes, that's a horrible little hack. :)

I think that it's going to be tricky to get a universal material
description language together -- but if you can do it, that's great.
My suggestion would be to go with a 'dictionaries' approach, and let the
user supply per output-system definitions in the preferred form for that
system.

e.g.,

"Three should use { ... } for 'oyster-blue-paper'"
"Cameo should use { pressure: 12 } for 'oyster-blue-paper'"

etc.

wdyt?

PR #63, opened 2019-04-05
Add JSxCAD.js to the .gitignore file

BarbourSmith:
Since we are trying not to check in the generated file app/v1/dist/.JSxCAD.js I think it makes sense to put it in the .gitignore file so changes to it will automatically not be tracked


pentacular:
Thanks

PR #64, opened 2019-04-05
Add a download button

BarbourSmith:
This pull request adds a download button to the pages in the app. I'm not 100% convinced this is the best way to do it, but it's an exploration into how it could be done. 

image

## Changes to JSxCAD in this PR

Adding a download button required one structural change to JSxCAD which is that the function writeFileSyncBrowser() will now write the translator to the file object, and readFileSyncBrowser() now returns the whole file object instead of just file.data. 

I think these changes make sense, but I don't fully understand every way these functions are intended to be used so its a change worth thinking about.

## Changes to app in this PR

This PR adds a download button to each viewer page. If the contents of the viewer do not contain the necessary information to build a valid download the button is disabled. The download process is lazy, ie the file is not created until you click the button and request it.


sys/writeFileSyncBrowser.js
```diff
@@ -2,7 +2,14 @@ const { getFile } = require('./files');
 
 const writeFileSync = (path, data, options) => {
   const file = getFile(path);
+  const { translator } = options;
+
   file.data = data;
+
+  if (translator) {
```
BarbourSmith: I have a hunch there is a cool way to do this all in one or two lines. Do you know how that could be done?

app/v1/dist/index.html
```diff
@@ -159,9 +163,11 @@
         let stats;
         let mesh;
         let gui;
+        let downloadButton;
 
-        watchFile(path, ({ data }) => {
+        watchFile(path, (file) => {
```
pentacular: Suggest either:

watchFile(path, ({ data, translator }) =>

or:

const { data, translator } = file.

Since file itself doesn't seem to be used, I'd recommend the first option.

app/v1/dist/index.html
```diff
@@ -187,6 +193,14 @@
                                                       .onChange((value) => { dataset.mesh.visible = value; });
                             }
                           }
+                          
+                          if(file.translator){
```
pentacular: Let's be boring and put spaces after if

if (translator) {
  ...
} else {
  ...
}

app/v1/dist/index.html
```diff
@@ -225,6 +239,20 @@
           gui = new dat.GUI({ autoPlace: false });
 console.log(`QQ/gui/domElement: ${gui.domElement}`);
```
pentacular: Oops, could you remove this QQ message I forgot about (unless you're using it)

app/v1/dist/index.html
```diff
@@ -225,6 +239,20 @@
...
           viewerElement.appendChild(gui.domElement);
+          
+          //Add a download button to the page it TODO: this process will need to updated for downloading other file types like .pdf
```
pentacular: How about we stick the downloadButton adding code in a little function, so that it's clear where the boundary of the code is?

sys/writeFileSyncBrowser.js
```diff
@@ -2,7 +2,14 @@ const { getFile } = require('./files');
 
 const writeFileSync = (path, data, options) => {
   const file = getFile(path);
+  const { translator } = options;
```
pentacular: This will break if options is undefined, so you will want

const writeFileSync = (path, data, options = {}) => {

above

sys/writeFileSyncBrowser.js
pentacular: How about

const writeFileSync = (path, data, { translator } = {}) => {
  const file = getPath(path);
  file.data = data;
  file.translator = translator;

I guess you could try using Object.assign(file, { data, translator }) but that's getting a bit hard to read, maybe.

app/v1/dist/index.html
```diff
@@ -225,6 +239,20 @@
...
+          
+          downloadButton = document.createElement("button");
+          downloadButton.setAttribute("type", "button");
+          downloadButton.appendChild(document.createTextNode("Download"));
+          downloadButton.setAttribute("style", "position: absolute; top: 2px; right: 2px;");
+          downloadButton.setAttribute("disabled", "true");
+          downloadButton.addEventListener("click", (e) => {
```
pentacular: If you don't use the (e) it's probably better to write () => { instead of (e) => {

app/v1/dist/index.html
```diff
@@ -225,6 +239,20 @@
...
+            const blob = new Blob([readFileSync(path).translator()], {type: "text/plain;charset=utf-8"});
```
pentacular: Ok, so we have one download button per page which displays stuff.

Looks good :)

sys/writeFileSyncBrowser.js
BarbourSmith: I like that. It's clear to read and simple

sys/readFileSyncBrowser.js
```diff
@@ -1,5 +1,5 @@
 const { getFile } = require('./files');
 
-const readFileSync = (path, options) => getFile(path).data;
+const readFileSync = (path, options) => getFile(path);
 
 module.exports.readFileSync = readFileSync;
```
pentacular: The readFileSync change will break on node.

Otherwise looks good.

BarbourSmith: So if I understand correctly right now we're in a spot where as long as a path is written to from the browser and then read from the browser everything is OK. If a path is written to from node and then read from node everything is OK. But if we were to somehow cross them then things would break, right? In theory that's not something which will happen, but it still seems wrong to have a function with the same name do different things in the browser and in node.

What do you think the solution is? Should I modify the node code to write an object with a translator to the fs? That doesn't seem like what we want...

pentacular:
l think readFileSync needs to get the data rather than a file object.

I think it's a trivial fix :)

pentacular:
Ah. I think I see the problem. :)

pentacular:
Yeah, it needs to return the application of the translator.

pentacular:
Actually, let me fix file a bit. It shouldn't be overloading data like that.

Data should always be the file data (or a lazy constructor for it).

The geometry can be exposed just to watchers

Sorry about that.

BarbourSmith:
Sounds like a plan :)

pentacular:
https://github.com/jsxcad/JSxCAD/pull/66

Will merge in an hour or two unless there's an objection. :)

But i think it does what you want.

BarbourSmith:
This should be good to go now with the changes integrated. Again, I like how you did them 👍  👍

sys/readFileSyncBrowser.js
```diff
@@ -1,5 +1,5 @@
 const { getFile } = require('./files');
 
-const readFileSync = (path, options) => getFile(path).data;
+const readFileSync = (path, options) => getFile(path);
```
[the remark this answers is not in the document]
pentacular: But I realize now that I didn't force lazy data on reading -- I'll put in a PR to fix that now.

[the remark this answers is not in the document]
pentacular: I think that merging and reverting the change to readFileSyncBrowser should be sufficient for everything to work.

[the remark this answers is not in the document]
BarbourSmith: I will do that right now. I think my biggest contribution right now is to break things in ways that you didn't expect 😛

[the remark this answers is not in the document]
pentacular: Well, it's understandable given that the fs abstraction was arse-about-face. ;)

app/v1/dist/index.html
```diff
@@ -222,9 +226,23 @@
...
           addPage(viewerElement);
           window.addEventListener( 'resize', onWindowResize, false );
         }
+        function makeDownloadButton() {
+          downloadButton = document.createElement("button");
+          downloadButton.setAttribute("type", "button");
+          downloadButton.appendChild(document.createTextNode("Download"));
+          downloadButton.setAttribute("style", "position: absolute; top: 2px; right: 2px;");
+          downloadButton.addEventListener("click", () => {
+            const blob = new Blob([readFileSync(path).data()], {type: "text/plain;charset=utf-8"});
```
pentacular: I'm pretty sure you need to remove the .data() from here, too. :)

BarbourSmith: Indeed I do 😬

pentacular:
Thanks.

FFR, x + "s" would be generally better as `${x}s` :)

PR #66, opened 2019-04-06
Refactor file interface to keep data and geometry consistently partitioned.

pentacular:


app/v1/dist/index.html
```diff
@@ -119,13 +117,13 @@
               cube({ size: 1, center: true }).translate([0.5, 0.5, 0.5]).as('b'),
               cube({ size: 1, center: true }).translate([0.5, 0, 0]).as('c').material('metal'))
 
-writePaths({ path: 'assembly' },
-           x.toPaths({ tags: ['a'] }),
-           x.toPaths({ tags: ['b'] }),
-           x.toPaths({ tags: ['c'] }))
+writeStl({ path: 'assembly' },
+         x.toSolid({ tags: ['a'] }),
```
BarbourSmith: I like this notation better, it makes more sense to me at a glance

sys/writeFileSyncBrowser.js
```diff
@@ -1,10 +1,10 @@
 const { getFile } = require('./files');
 
-const writeFileSync = (path, data, options) => {
+const writeFileSync = (path, data, options = {}) => {
   const file = getFile(path);
   file.data = data;
   for (const watcher of file.watchers) {
-    watcher(file);
+    watcher(file, options);
```
BarbourSmith: 👍

BarbourSmith:
Looks good to me 👍  👍

app/v1/dist/index.html
pentacular: Cool.

Although I hope that the x.toSolid() stuff can be made mostly invisible to end-users at some point by adding some nice API support. :)

PR #68, opened 2019-04-08
Add A writeSvg() function and expose it through the api

BarbourSmith:
This pull request adds a writeSvg method to the api which works similar to the way writePdf does. Most of the heavy lifting was already done so this pull request basically just exposes the function in a usable way through the api.

There is currently an issue which is that circles do not appear entirely within the viewBox. I believe that the issue is with measureBoundingBox() but because that function is used in many places I think any changes to it should be made in a separate pull request.


example/v1/cutCubes.test.js
```diff
@@ -4,7 +4,7 @@ import { main } from './cutCubes';
 
 main();
 
-test('Expected stl', t => {
+test('Expected pdf', t => {
```
pentacular: Thanks :)

BarbourSmith:
I knew I had seen you handle default options in a better way somewhere, but I couldn't find it. Thank you 🙂

pentacular:
Thanks :)

PR #73, opened 2019-04-11
Expose bounding box function

BarbourSmith:
Add a .getBounds function to CSG in the style of jscad


api/v1/CSG.js
```diff
@@ -20,6 +20,10 @@ export class CSG {
     return CSG.fromGeometry(this.geometry.difference(...shapes.map(toGeometry)));
   }
 
+  getBounds () {
+    return measureBoundingBox([this.toPoints({})]);
```
pentacular: I think I gave some bad advice :)

What you're doing here is taking the points of an object and treating it as a polygon, which it isn't. :)
This will work because measureBoundingBox doesn't consider it as a polygon.

I've just added algorithms-points to address this.

import { measureBoundingBox } from '@jsxcad/algorithm-points';

then

return measureBoundingBox(this.toPoints())

should probably do what you want.

You'll also need to add algorithm-points as a dependency and then "npm run bootstrap" in the root directory for it to take effect.

Sorry about that.

pentacular:
Thanks :)

PR #123, opened 2019-05-14
Refactor Shape for readability.

pentacular:


BarbourSmith:
I like this name change a lot

PR #124, opened 2019-05-15
Generate diagrams for documentation

BarbourSmith:
I am starting the process of generating diagrams for the user guide [here](https://github.com/jsxcad/JSxCAD/wiki/User-Guide)

Before I create them all I wanted to check in to see if you have any feedback or preference about how it is done. Does this seem like the right approach? Any changes you would like to see before I do the rest?


BarbourSmith:
Note if you want to see the actual diagrams instead of the raw svg click the button rich diff button in the upper right hand corner

image

pentacular:
I expect this'll be an evolutionary process, and this looks like a good start.

Let's give it a try. :)

pentacular:
Ah. Some lintery to do.

PR #140, opened 2019-05-23
Update evaluator return and user guide

pentacular:


api/v1-bundle/webworker.js
```diff
@@ -9,12 +9,12 @@ const agent = async ({ ask, question }) => {
...
+      console.log(`QQ/script: ${question.evaluate}`);
+      console.log(`QQ/ecmascript: ${ecmascript}`);
       const code = new Function(`{ ${Object.keys(api).join(', ')} }`, ecmascript);
       const shape = await code(api).main();
       if (shape !== undefined) {
-        await sys.writeFile({ preview: true, geometry: shape.toDisjointGeometry() }, 'preview', 'preview');
+        return shape.toDisjointGeometry();
       }
     }
   } catch (error) {
```
BarbourSmith: I think that the catch statement here might need to be updated to reflect the new style. If that's the case, sorry for not noticing earlier 🙂

pentacular: That should be correct for writing to the log.
If you want, we can add an option to disable the log channel.

If you want to return an error message along the main answer path, we can also do that.

PR #141, opened 2019-05-23
Add option for camera near and far view

BarbourSmith:
I was running into an issue when converting a large item to a .svg image where it was too big to still be visible once I backed the camera up enough to see it all 🙄.

This should break out options for setting the camera view max and min.

Before merging will you double check that this doesn't cause any issues for the examples generation? I wanted to double check, but I wasn't sure where the script to generate the docs is


convert/threejs/toSvg.js
```diff
@@ -10,10 +10,10 @@ const { Projector, RenderableFace, RenderableLine, RenderableSprite } = installP
 const { SVGRenderer } = installSVGRenderer({ THREE, Projector, RenderableFace, RenderableLine, RenderableSprite, document: new DOMParser().parseFromString('<xml></xml>', 'text/xml') });
 // Bootstrap done.
 
-const build = ({ view = {}, pageSize = [100, 100], grid = false }, geometry) => {
+const build = ({ view = {}, pageSize = [100, 100], grid = false, near = 1, far = 3500 }, geometry) => {
   const { target = [0, 0, 0], position = [40, 40, 40], up = [0, 0, 1] } = view;
```
pentacular: Could you put these options into 'view' together with target and position?

PR #164, opened 2019-06-06
Rebase square, circle, triangle on polygon. Support apothem.

pentacular:


BarbourSmith:
Nice! Using polygon as a base makes a lot of sense

pentacular:
Thinking of doing likewise with prism for solids.

BarbourSmith:
That makes sense to me. Can all of the solids be covered by prism?

pentacular:
No :)

Cube and cylinder can be.

Tetrahedron, sphere, etc, not so much.

PR #169, opened 2019-06-08
Add temporary TJunction fix, wireframe, etc.

pentacular:


pentacular:
Note that this will make writing STL incredibly slow until re-implemented, but cursory examination indicates no severe defects with this approach.

PR #200, opened 2019-07-02
Build process integration 

BarbourSmith:
This pull request adds a new file called maslowWorker.js which bundles up the worker for my version. Let me know if there are ways I can clean this up or make it better.


BarbourSmith:
I have tried to reset master.js and webworker.js to the state they were in before I touched them, but I have still made some changes somehow so it is probably worth noting that and regenerating those files.

pentacular:
Ah, either need to clean up maslowWorker or add a .eslintignore for it.

PR #203, opened 2019-07-04
Export allTags

pentacular:


BarbourSmith:
From a look through the code (on a phone screen) I'm not sure I understand exactly how this works, but I am excited by what I can understand

pentacular:
Should traverse the shape geometry, building up a set of distinct tags.

PR #245, opened 2019-08-22
Add hud display and plan geometry

pentacular:


BarbourSmith:
This sounds cool, what exactly does the HUD camera do? Is it an overview of the whole project when you are working close in?

PR #286, opened 2019-09-21
Clean up tool paths and make them return a continuous path

BarbourSmith:
This pull request isn't ready to be merged, but I would love any feedback you have on the direction it is moving in.

This pull request makes it so that tool paths are continuous which will make the next step of deriving gcode from them much easier. 

image

However this pull request currently breaks the corner overcut behavior. I've been thinking about how that should work.

The way most CAD programs handle corner over-cuts is to subtract a cylinder on the corner. The worst but easiest way to do that is to place the cylinder directly on the corner.

image

The more nuanced approach I have seen is to set the edge of the circle on the corner so that no more material is removed than necessary like so:

image

That solution seems optimal to me for right angles, but doesn't properly handle the corner case of very small angles. You can see that in this case placing a circle tangent to the tip creates a shape which is still un-manufacturable because the tool can't access that circle.

image

What we need to do is to add a spur to the tool path which extends the complete path down until it is tangent to the corner. That will guarantee a shape which can be manufactured, which I haven't seen any CAD program do before. Your thinking about the path based approach I think is spot on.

What I think we need to do is add a little spur to our path which bisects the angle between the two paths and extends to within one bit diameter of the corner like so:

image


pentacular:
Very nice.

Just one request - could you make this optional in some way?

BarbourSmith:
Absolutely. I will add a option to turn it on if needed. Something like {joinPaths = false} by default?

pentacular:
Let's follow uo

PR #392, opened 2019-12-05
Sync up with JSxCAD Master

BarbourSmith:
I know this is about to be obsolete, but I figure we might as well sync up right before that anyway for good record keeping


BarbourSmith:
Oops, gotta lint

BarbourSmith:
I'm seeing the CI fail with:

image

But I am unsure what to do to make it happier

pentacular:
It's due to:

Missing dependencies
* @jsxcad/cache: ./maslowWorker.js

BarbourSmith:
Thanks!

PR #499, opened 2020-01-22
Update to keep things in sync with the latest syntax

BarbourSmith:
This pull request also adds the change to make the rollup process work


pentacular:
Thanks.

Could you "npm run lintfix" or add a .eslintignore entry for maslowWorker.js ?

Otherwise travis won't pass it. :)

BarbourSmith:
I woke up in the middle of the night and realized I forgot to run lint 😁. Thanks for the reminder.

I also realized that assemble to Assembly hasn't been changed so I might need another pass to clean things up. In general I think I am importing some things which are available in the API so a general cleanup of my imports is probably in order once I get the worker running.

pentacular:
npm run depcheck

should help with sorting out imports.

ui/v1/maslowWorker.js
```diff
@@ -46,19 +47,23 @@ const agent = async ({ ask, question }) => {
...
         const returnVal = foo({ ...api, ...inputs });
-        return returnVal.toDisjointGeometry();
+        if (typeof returnVal === 'object') {
+          return returnVal.toDisjointGeometry();
+        } else {
+          return returnVal;
+        }
       case 'getLayoutSvgs':
         // Extract shapes
         let items = api.Shape.fromGeometry(values[0]).toItems();
```
pentacular: You might want to try unpacking a bit like this ...

case 'getLayoutSvgs': {
  const [g, margin, width, length] = values;
  ...
}

ui/v1/maslowWorker.js
```diff
@@ -4,15 +4,16 @@ import * as api from '@jsxcad/api-v1';
...
 import * as sys from '@jsxcad/sys';
 
-import { assemble, difference, intersection, union } from '@jsxcad/api-v1-shape';
+import { assemble, intersection, union } from '@jsxcad/api-v1-shape';
 import { clearCache } from '@jsxcad/cache';
 import { hull } from '@jsxcad/api-v1-extrude';
-import { pack } from '@jsxcad/api-v1-layout';
+// import { pack } from '@jsxcad/api-v1-layout';
 import { toStl } from '@jsxcad/convert-stl';
 import { toSvg } from '@jsxcad/convert-svg';
```
pentacular: It occurs to me that you might write something like

const shape = geometry => api.Shape.fromGeometry(geometry); // Since this gets written out a lot.

const interface = {
  'bounding_box': g => shape(g).measureBoundingBox(),
  'circle': (diameter, sides) => Circle.ofDiameter(diameter, { sides }).toDisjointGeometry(),
  ... //
}

Then call

return (interface(key) || error)(...values);

would probably do the trick

PR #745, opened 2020-08-17
Add workerpool support for maslowWorker.js

pentacular:


BarbourSmith:
Thank you!

Were all of these changes needed just to add worker support?

pentacular:
I think that the last push didn't fully rebuild the notebooks, so they got updated on this rebuild.

PR #848, opened 2020-12-23
Display grid even if fit = false

BarbourSmith:
I am not 100% sure this is the right way to do this, but I figured better to at least try to make a pull request than ask you to change it. 

Basically if the fit setting was turned off in an orbitDisplay then the moveToFit function was returning before the grid was generated. This change moves that return until after the grid has been created.


pentacular:
Ah, very nice -- thanks. :)

PR #859, opened 2021-01-14
Expose grid layer

BarbourSmith:
This pull request exposes the grid layer. Some changes to maslowWorker slipped in there too.

How do you feel about me removing maslowWorker entirely? At this point I am not really using it since the new bundling system exposes things in an easy to use way which doesn't require maslowWorker to be bundled with everything else.


pentacular:
Thanks

PR #959, opened 2021-07-08
Tweaks to orbitView

BarbourSmith:
A couple tweaks to orbit view to make it easier to turn on and off the background grid and reset the camera to the original position

### DIFF
--- a/ui/threejs/moveToFit.js
+++ b/ui/threejs/moveToFit.js
@@ -68,6 +68,8 @@ export const moveToFit = ({
     return;
   }
 
+   controls.reset();
+
   const center = box.getCenter(new Vector3());
   // const size = box.getSize(new Vector3());
 
--- a/ui/threejs/moveToFit.js
+++ b/ui/threejs/moveToFit.js
@@ -68,7 +68,7 @@ export const moveToFit = ({
     return;
   }
 
-   controls.reset();
+  controls.reset();
 
   const center = box.getCenter(new Vector3());
   // const size = box.getSize(new Vector3());
--- a/ui/threejs/orbitDisplay.js
+++ b/ui/threejs/orbitDisplay.js
@@ -94,7 +94,7 @@ export const orbitDisplay = async (
 
   const updateGeometry = async (
     geometry,
-    { withGrid = true, fit = true} = {}
+    { withGrid = true, fit = true } = {}
   ) => {
     // Delete any previous dataset in the window.
     for (const { mesh } of datasets) {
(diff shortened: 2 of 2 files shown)


ui/threejs/moveToFit.js
```diff
@@ -68,6 +68,8 @@ export const moveToFit = ({
     return;
   }
 
+   controls.reset();
```
BarbourSmith, line 71: This resets the camera to point back at the center if it has been panned. This hasn't been an issue since we are creating a new orbitView with each view, but when using updateGeometry the camera would zoom to fit everything but be pointed at the wrong spot.

ui/threejs/orbitDisplay.js
```diff
@@ -92,12 +92,17 @@ export const orbitDisplay = async (
     render();
   }).observe(page);
 
-  const updateGeometry = async (geometry) => {
+  const updateGeometry = async (
+    geometry,
+    { withGrid = true, fit = true} = {}
```
BarbourSmith, line 97: This exposes withGrid and fit in updateGeometry. I tried to make it so that it wouldn't impact the way preAlpha works,  but I'm not 100% sure that I did that correctly so it's worth checking that I didn't break something here.

ui/threejs/orbitDisplay.js
```diff
@@ -92,12 +92,17 @@ export const orbitDisplay = async (
...
+  ) => {
     // Delete any previous dataset in the window.
     for (const { mesh } of datasets) {
       scene.remove(mesh);
     }
 
+    view.fit = fit;
```
pentacular, line 104: Rather than modifying the incoming structure, how about using { ...view, fit } instead of view, below?

ui/threejs/orbitDisplay.js
pentacular: Looks good

pentacular: Can you make the spacing happy above?

BarbourSmith: Did I fix the spacing in the next commit for lint?

ui/threejs/orbitDisplay.js
```diff
@@ -92,12 +92,17 @@ export const orbitDisplay = async (
     render();
   }).observe(page);
 
-  const updateGeometry = async (geometry) => {
+  const updateGeometry = async (
+    geometry,
+    { withGrid = true, fit = true } = {}
```
BarbourSmith, line 97: This is the new spacing

PR #980, opened 2021-08-06
Fix preact state delay

pentacular:
Looks like preact doesn't propagate state changes immediately, so we need to advise the function directly in some cases.

Also seem to have fixed the widget size computation problem.


api/shape/view.js
```diff
@@ -18,8 +18,8 @@ export const baseView =
     wireframe = false,
     prepareView = (x) => x,
     inline,
-    width = 1024,
```
BarbourSmith, line 21: Does this make views smaller by default?

es6/jsxcad-ui-notebook.js
```diff
@@ -2767,10 +2767,12 @@ const toDomElement = (notebook = [], { onClickView } = {}) => {
       Object.assign(entry, note.define.data);
     }
     if (note.view) {
-      const { url, openView } = note;
+      const { url, openView, view } = note;
+      const { height, width } = view;
       const image = document.createElement('img');
       image.style.display = 'block';
-      image.style.height = `${21 * 13}px`;
+      image.style.height = `${height}px`;
```
BarbourSmith, line 2774: Nice!

ui/v1/ui.js
```diff
@@ -304,6 +305,8 @@ class Ui extends PureComponent {
             def.domElement.appendChild(document.createTextNode(entry.hash));
             def.domElement.appendChild(element);
             console.log(`Marking ${entry.hash} in ${id}`);
+            // Let the node attach.
+            await sleep(0);
```
BarbourSmith, line 309: How does this work? Is the point that sleep is non-blocking here so that thread is able to handle other calls on the stack?

BarbourSmith (approved):
Looks good to me 🙂

api/shape/view.js
pentacular: Yes, but I think they were rendering smaller anyhow -- when I fixed the sizes they got very large. :)

ui/v1/ui.js
pentacular: Ah, I thought I'd removed that.

The idea was to let the browser render the dom nodes -- the active code is in

      await new Promise((resolve, reject) => {
        requestAnimationFrame(resolve);
      });

in JsEditor

pentacular: Without this we get a zero offsetHeight on the widget element.

PR #982, opened 2021-08-07
More state synchronization

pentacular:


ui/v1/schedule.js
```diff
@@ -0,0 +1,7 @@
+/* global requestAnimationFrame */
+
+export const animationFrame = () => {
```
BarbourSmith, line 3: This is nice and clean as a stand alone function now

BarbourSmith (approved):

PR #983, opened 2021-08-07
Update html generator and notebook testing

pentacular:


ui/threejs/mesh.js
```diff
@@ -13,7 +13,7 @@ import {
   Vector2,
   Vector3,
   VertexColors,
-} from 'three';
+} from 'three/build/three.module.js';
```
BarbourSmith, line 16: Why the switch to more explicit imports in these?

BarbourSmith (approved):

ui/threejs/mesh.js
pentacular: Good catch -- I did this initially because there was an issue with importing them into node, but that was factored out.

I'll revert these.

PR #984, opened 2021-08-08
Update html and image generation from notebooks with tests against expectations

pentacular:


nb/projects/gear/gear.nb
```diff
@@ -19,5 +19,4 @@ const example = Gear()
   .hasToothResolution(toothResolution)
   .ex(thickness)
   .material('wood')
-  .view(1024)
```
BarbourSmith, line 22: I know this is a deleted line, but what does the argument to view() do?

BarbourSmith (approved):
I don't have any suggestions for changes, but I did learn some things by reading through and I'm even more impressed with this system now than I was before. Looks good to me.

nb/projects/gear/gear.nb
pentacular: It was a size at one time. :)

BarbourSmith: Ah right, I remember that now.

PR #1003, opened 2021-08-18
Add fit/fitTo, clean up Hershey, support functors in and, cut, etc.

pentacular:


api/shape/Hershey.js
```diff
@@ -1431,10 +1431,8 @@ export const toPaths = (letters) => {
     .outline();
 };
 
-export const ofSize = (size) => (text) => toPaths(text).scale(size);
+export const ofSize = (text, size) => toPaths(text).scale(size);
```
BarbourSmith, line 1434: Does this change how Hershey works? How would I use it after the change?

api/shape/main.js
```diff
@@ -280,7 +281,9 @@ export { minkowskiSum } from './minkowskiSum.js';
 export { move } from './move.js';
 export { noVoid } from './noVoid.js';
 export { notAs } from './notAs.js';
+export { nth } from './nth.js';
```
BarbourSmith, line 284: It's fun to see these get added 👍 👍

BarbourSmith (approved):
Looks good to me

api/shape/Hershey.js
pentacular: Ah, it changes it to match how Text() works.

Hershey(12)('Hello') becomes Hershey('Hello', 12)

PR #1008, opened 2021-08-25
Documentation proposal

BarbourSmith:
I was messing around with some ideas for how to structure the documentation. What are your thoughts on this organization structure?

This isn't complete, but it feels like it's grown beyond copying and pasting in the google doc.


pentacular:
Looks good :)

I think we need to rename those .md files as .nb.

The build process broke because the observed start.md differed from the expected start.md.
You could fix that by running doc locally and then copying start.observed.md to start.md after checking the diff was intended.

That would also produce .md files for each of the .nb files and .md.0.png, etc, which would be checked in.

But you don't need to worry about that if it's too expensive -- I'll can fix those up in a later PR.

Let's rename the .md as .nb, and how about separating the words with _ rather than just sticking them together?

e.g., foodoc.nb -> foo_doc.nb

See if 'npm run doc' works after that, but don't worry about it if it takes too long or has problems -- but it would be good to know if it has issues under windows. :)

BarbourSmith:
Name changes are in, but I'm running into:

C:\Users\Bar\Documents\GitHub\JSxCAD4>npm run doc

```
> @jsxcad/core@1.0.0 doc C:\Users\Bar\Documents\GitHub\JSxCAD4
> node doc/update.js

internal/process/esm_loader.js:74
    internalBinding('errors').triggerUncaughtException(
                              ^

Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './dist/cjs-compat' is not defined by "exports" in C:\Users\Bar\Documents\GitHub\JSxCAD4\node_modules\idb-keyval\package.json imported from C:\Users\Bar\Documents\GitHub\JSxCAD4\sys\db.js
    at throwExportsNotFound (internal/modules/esm/resolve.js:289:9)
    at packageExportsResolve (internal/modules/esm/resolve.js:534:3)
    at packageResolve (internal/modules/esm/resolve.js:665:14)
    at moduleResolve (internal/modules/esm/resolve.js:716:14)
    at Loader.defaultResolve [as _resolve] (internal/modules/esm/resolve.js:809:11)
    at Loader.resolve (internal/modules/esm/loader.js:85:40)
    at Loader.getModuleJob (internal/modules/esm/loader.js:229:28)
    at ModuleWrap.<anonymous> (internal/modules/esm/module_job.js:51:40)
    at link (internal/modules/esm/module_job.js:50:36) {
  code: 'ERR_PACKAGE_PATH_NOT_EXPORTED'
}
```
when running 'npm run doc'

pentacular:
Hmm, that's an interesting error.

I see this article, which may be relevant.

https://stackoverflow.com/questions/62246824/error-err-package-path-not-exported-no-exports-main-resolved-in-app-node-m

That looks great, by the way -- I think we can just merge it in if npm run doc is being a problem.

PR #1010, opened 2021-09-01
Switch to a replay/update/export eval model.

pentacular:
This changes the evaluation model to a single, out-of-order pass.

Initially pre-computed replays are replayed, followed by updates in a dependency ordering.
Finally exports are computed and returned.

This will need a bit more work to handle imports efficiently -- probably by adding an import pass prior to updates.


pentacular:
This is a bit of a transitional point -- there's some more work to do to clean up compilation.

So I haven't fully updated the compiler tests.

This also updates the nb/documentation which was merged in.

api/core/evaluate.js
```diff
@@ -28,82 +27,74 @@ export const execute = async (
...
-    const processed = new Set();
-    let somethingHappened;
-    let somethingFailed;
-    let parallelUpdates = 0;
-    const schedule = () => {
-      console.log(`Updates remaining ${[...pending].join(', ')}`);
-      for (const id of [...pending]) {
-        if (parallelUpdates >= parallelUpdateLimit) {
-          break;
+    let replaysDone = false;
```
BarbourSmith, line 30: Is this the global state that you would like to eliminate?

api/core/evaluate.js
```diff
@@ -28,82 +27,74 @@ export const execute = async (
...
-    while (unprocessed.size > 0) {
-      const somethingHappens = new Promise((resolve, reject) => {
-        somethingHappened = resolve;
-        somethingFailed = reject;
-      });
-      schedule();
-      if (unprocessed.size > 0) {
-        // Wait for something to happen.
-        await somethingHappens;
+      // FIX: We could instead use Promise.race() and then see what new updates could be queued.
```
BarbourSmith, line 85: Promise.race() is a neat feature that I didn't know about. That's a good one to know.

build-notest.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+npm run bootstrap && npm run doc && npm run depcheck && npm run prettier && npm run lint && npm run rollup && ./publish-es6.sh && ./publish-wasm.sh && echo 'DONE'
```
BarbourSmith, line 3: Is this an overall build script?

compiler/toEcmascript.test.js
```diff
@@ -192,8 +192,8 @@ popSourceLocation({
 return {};
 
 } catch (error) { throw error; }
-`
-  );
+`,
+  ]);
 });
 
 test('Bind await to calls properly.', async (t) => {
```
BarbourSmith, line 199: What's going on here with the aaaaa? 😛

nb/documentation/creating_geometry.md
```diff
@@ -0,0 +1,100 @@
+# Creating Geometry
```
BarbourSmith, line 1: Thanks for generating these 🙂

BarbourSmith (approved):
This is a big pull request with a lot of work in it. I can't say that I fully understand every intricacy of the replay system, but I certainly understand it better than if I hadn't looked at it so thanks for letting me read through it.

build-notest.sh
pentacular: build.sh rebuilds to generate a PR that should pass.
build-notest.sh does the same, but skips the test step -- the regression notebooks are run.

compiler/toEcmascript.test.js
pentacular: There was some weird issue about excessively long lines, I think.

nb/documentation/creating_geometry.md
pentacular: np

PR #1012, opened 2021-09-07
Update url generation to support windows

BarbourSmith:
Only tested on windows so far


algorithm/cgal/getCgal.js
```diff
@@ -1,4 +1,5 @@
 import { emit, isNode, log, onBoot } from '@jsxcad/sys';
+import { fileURLToPath } from 'url';
```
BarbourSmith, line 2: Is this in the right order?

PR #1014, opened 2021-09-08
Update compiler to import/replay/update/export model

pentacular:


pentacular:
I redid the getCgal code to use URL so that it would rollup for the browser -- could you check this actually works under windows?

nb/documentation/file_import_and_export.md
```diff
@@ -1,2 +1,26 @@
 # File IO
-TODO: This needs to be finished.
+JSxCAD supports importing and exporting .svg and .stl file types
```
BarbourSmith, line 2: Thanks for generating these

algorithm/cgal/getCgal.js
```diff
@@ -33,9 +32,10 @@ export const initCgal = async () => {
         if (path === 'cgal_node.wasm' || path === 'cgal_browser.wasm') {
           let url = import.meta.url;
           if (url.startsWith('file://')) {
-            url = fileURLToPath(url);
-            url = url.substring(0, url.length - 10) + 'cgal_node.wasm';
-            return url;
+            const { pathname } = new URL(url);
```
BarbourSmith, line 35: I think this is resulting in the issue with the leading '\' again. I'm seeing "/C:/Users/Bar/Documents/GitHub/JSxCAD4/algorithm/cgal/cgal_node.wasm" as the generated path

PR #1021, opened 2021-09-22
Add loop and loft to documentation

BarbourSmith:
The loft example is pretty weak, but I had a hard time figuring out how to do two operations at each step (ie move up and shrink).

I also need to update the notebooks so this passes the tests. After I run 'npm run doc' which files need to go from where to where again?


pentacular:
> The loft example is pretty weak, but I had a hard time figuring out how to do two operations at each step (ie move up and shrink).
> 
> I also need to update the notebooks so this passes the tests. After I run 'npm run doc' which files need to go from where to where again?

At the end of the npm run doc output you should see something like

Expectations failed:
difference: cp nb/documentation/interactions_with_geometry/interactions_with_geometry.observed.md nb/documentation/interactions_with_geometry/interactions_with_geometry.md
difference: display nb/documentation/interactions_with_geometry/interactions_with_geometry.md.8.difference.png
            cp nb/documentation/interactions_with_geometry/interactions_with_geometry.md.8.observed.png nb/documentation/interactions_with_geometry/interactions_with_geometry.md.8.png

The cp commands are what should make the test pass next time, although you may also need some git add for new files.

pentacular:
> The loft example is pretty weak, but I had a hard time figuring out how to do two operations at each step (ie move up and shrink).

Ah, that is non-obvious.

Here's an example which shows the composition more clearly.

Triangle(10)
  .cut(Triangle(5))
  .loft(
    ...seq(
      (a) => (s) =>
        s
          .rz(a / 5)
          .scale(a)
          .z(a * 10),
      { by: 1 / 8 }
    )
  )
  .view();

Does that make sense?

pentacular:
Let me know if you want this merged as, or if you want to adjust things a bit. :)

BarbourSmith:
Thanks for the loft example, it does make sense. Would it be possible for seq() to accept a single function as an argument? It seems to me that it would be more intuitive.

I'm closer, but I'm still running into an issue when I go to build the docs. I'm hitting an error message when I run 'npm run docs' in documentation.

```
QQ/Evaluate
QQ/Compile
Read complete: control/nb\documentation\user_interface\user_interface.nb missing
Read complete: meta/def/nb\documentation\user_interface\user_interface.nb/$1 missing
QQ/acquire
Failed with error
SyntaxError: Bad character escape sequence (1:46)
```

Looking at the user_interface.nb file it's pretty much empty with no obvious weird characters. What are your thoughts?

pentacular:
> Thanks for the loft example, it does make sense. Would it be possible for seq() to accept a single function as an argument? It seems to me that it would be more intuitive.

It could, but I'm not sure what the stepping defaults would be.

e.g., seq(rx) ?

Or do you mean removing the second argument to the function?

So that seq(rx, { by: 1/8 }) doesn't call rx(1/8, 0), rx(2/8, 1), etc?
(This should be merged in now, I think)

> 
> I'm closer, but I'm still running into an issue when I go to build the docs. I'm hitting an error message when I run 'npm run docs' in documentation.
> 
> ```
> QQ/Evaluate
> QQ/Compile
> Read complete: control/nb\documentation\user_interface\user_interface.nb missing
> Read complete: meta/def/nb\documentation\user_interface\user_interface.nb/$1 missing
> QQ/acquire
> Failed with error
> SyntaxError: Bad character escape sequence (1:46)
> ```
> 
> Looking at the user_interface.nb file it's pretty much empty with no obvious weird characters. What are your thoughts?

Odd -- let me see if I can try it locally.

pentacular:
> > Thanks for the loft example, it does make sense. Would it be possible for seq() to accept a single function as an argument? It seems to me that it would be more intuitive.
> 
> It could, but I'm not sure what the stepping defaults would be.
> 
> e.g., seq(rx) ?
> 
> Or do you mean removing the second argument to the function?
> 
> So that seq(rx, { by: 1/8 }) doesn't call rx(1/8, 0), rx(2/8, 1), etc?
> (This should be merged in now, I think)
> 
> > I'm closer, but I'm still running into an issue when I go to build the docs. I'm hitting an error message when I run 'npm run docs' in documentation.
> > ```
> > QQ/Evaluate
> > QQ/Compile
> > Read complete: control/nb\documentation\user_interface\user_interface.nb missing
> > Read complete: meta/def/nb\documentation\user_interface\user_interface.nb/$1 missing
> > QQ/acquire
> > Failed with error
> > SyntaxError: Bad character escape sequence (1:46)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Looking at the user_interface.nb file it's pretty much empty with no obvious weird characters. What are your thoughts?
> 
> Odd -- let me see if I can try it locally.

Ok, I've run it locally and it works as expected (this matches the CI logs as well).

I guess it must be an issue at your end -- which version of node are you running with?

PR #1022, opened 2021-09-22
Simpler compiler

pentacular:
This delegates replay/compute decisions to run-time, and cleans up mutual exclusivity on evaluations.

We will consider optimizing where we know ahead of time that replay is possible in a follow-up.


api/v1-gcode/writeGcode.js
```diff
@@ -15,10 +14,7 @@ export const prepareGcode = (shape, name, tool, options = {}) => {
   let index = 0;
   const entries = [];
   for (const entry of ensurePages(shape.toKeptGeometry())) {
-    const op = toGcode(entry, tool, {
-      definitions: getDefinitions(),
-      ...options,
-    }).catch(getPendingErrorHandler());
+    const op = toGcode(entry, tool, options).catch(getPendingErrorHandler());
```
BarbourSmith, line 17: The change to these is that options is now an object? Like {option1: true}

BarbourSmith (approved):
Looks good to me :)

api/v1-gcode/writeGcode.js
pentacular: Ah, it already was -- the difference is that we aren't mixing definitions into it anymore.

PR #1027, opened 2021-09-30
Add abstract, clarify bom, add getNot, promote view / output op transform, update tags, replace compose/non-positive with type:void

pentacular:


BarbourSmith (approved):
👍 👍

PR #1029, opened 2021-10-03
Add faces()

pentacular:
This also fixes color vs tint and makes polygonsWithHoles more of a first class geometry.


api/shape/color.js
```diff
@@ -1,14 +1,7 @@
...
-export const fromName = (shape, name) =>
-  Shape.fromGeometry(rewriteTags(toTagsFromName(name), [], shape.toGeometry()));
-
-// FIX: Have color remove all other color tags.
-export const color =
-  (...args) =>
-  (shape) =>
-    fromName(shape, ...args);
+export const color = (name) => (shape) =>
+  shape.untag('color:*').tag(...toTagsFromName(name));
```
BarbourSmith, line 5: This came together neatly

BarbourSmith (approved):
Looks good to me :)

api/shape/color.js
pentacular: Yes, gradually getting more civilized.

PR #1031, opened 2021-10-06
Expand gear.nb

pentacular:


es6/jsxcad-geometry.js
```diff
@@ -1900,17 +1900,12 @@ const grow = (geometry, amount) => {
...
-// This alphabet uses `A-Za-z0-9_-` symbols. The genetic algorithm helped
-// optimize the gzip compression for this alphabet.
 let urlAlphabet =
   'ModuleSymbhasOwnPr-0123456789ABCDEFGHNRVfgctiUvz_KqYTJkLxpZXIjQW';
-
 let nanoid = (size = 21) => {
   let id = '';
-  // A compact alternative for `for (var i = 0; i < step; i++)`.
   let i = size;
   while (i--) {
```
BarbourSmith, line 1908: This is interesting. Should I be doing this instead of `for (var i = 0; i < step; i++)`?

nb/projects/gear/gear.md
```diff
@@ -93,52 +95,162 @@ Gear()
...
-Arc(41)
-  .cut(gear.hasTeeth(32).hasBacklash(-0.5).hasClearance(-0.1))
-  .md(`Some care needs to be taken to invert play when cutting gears out.`)
+const ring = Arc(50)
+  .hasAngle(-1 / 64, 1 / 64)
+  .hull(Point())
+  .cut(Arc(30))
+  .cut(
+    Group(
+      ...seq(
```
BarbourSmith, line 121: sequence is turning out to be useful for all kinds of things

nb/projects/gear/gear.md
```diff
@@ -93,52 +95,162 @@ Gear()
...
+  .stl('solar', (s) => s.get('solar').getNot('axle'))
+  .stl('axle', (s) =>
+    s
+      .get('solar')
+      .get('axle')
+      .rx(0 / 2, 1 / 2)
+  );
 ```
 
-Some care needs to be taken to invert play when cutting gears out.
```
BarbourSmith, line 124: Not any more 😀 Your solution takes care of that nicely.

BarbourSmith (approved):
The new technique looks great to me 👍 👍

es6/jsxcad-geometry.js
pentacular: Well, not really, since it does something else.

As an alternative to for (let i = step; i > 0; i--) { ... } perhaps. :)

But I'd prefer the for statement on the grounds that it limits scope better.

PR #1035, opened 2021-11-01
Change bend and taper, and fix stl ordering.

pentacular:
Bend now takes a radius for the reference circle to fit the curve to.
Taper takes four factors of z.
Stl should have triangles leading with the minimal vertex.


api/shape/define.js
```diff
@@ -14,73 +14,69 @@ export const defThreejsMaterial = (name, definition) =>
...
 
-export const defGrblDynamicLaser = (
-  name,
-  {
-    cutDepth = 0.2,
-    diameter = 0.09,
-    jumpPower = 0,
-    power = 1000,
-    speed = 1000,
+export const GrblDynamicLaser = ({
```
BarbourSmith, line 36: Are you thinking about doing some laser stuff?

nb/documentation/interactions_with_geometry/interactions_with_geometry.md
```diff
@@ -164,43 +164,47 @@ Remesh can be used to break up the segments of a shape allowing it to be distort
...
 
 ![Image](interactions_with_geometry.md.14.png)
 
+A rectangle bent into a ring with a central radius of 25
+
+---
+### Rotate
+A shape can be rotated about the origin using the .rx(), .ry(), and .rz() commands.
+These take the number of turns as an argument .rz(1/8) would rotate the shape by 1/8th of a rotation.
+Multiple turns can be provided, which will produce one result per turn.
```
BarbourSmith, line 185: That's neat, I wasn't aware of that feature

BarbourSmith (approved):
As always this looks great to me 😀

nb/documentation/interactions_with_geometry/interactions_with_geometry.md
pentacular: Yeah, the fly in the ointment is that there's no way to post-process on a per turn basis, but I think a parallel op might work.

e.g., s.rz(0/1, 1/2).op(color('blue'), color('red')) might work, except it needs something to distinguish between an op on the whole shape and an op on s.n(0), s.n(1), s.n(2).

api/shape/define.js
pentacular: Laser stuff was working ok, but this is just clean up.

I'll get the laser set up in a while and make sure it's working with the simplified system.

PR #1036, opened 2021-11-03
Add Surface.to

pentacular:


api/shape/to.js
```diff
@@ -0,0 +1,31 @@
...
+import { invertTransform } from '@jsxcad/algorithm-cgal';
+
+export const to =
+  (selection, ...ops) =>
+  (shape) => {
+    if (ops.length === 0) {
+      ops.push(() => shape);
+    }
+    ops = ops.map((op) => (op instanceof Function ? op : () => op));
+    // We've already selected the item for reference, e.g., s.on(g('plate'), ...);
```
BarbourSmith, line 13: These inline comments are really nice for understanding what's going on

BarbourSmith (approved):
👍 👍

api/shape/to.js
pentacular: Good :)

PR #1038, opened 2021-11-04
Add project to generate sprocket profiles for different belts and small tweak to .scale() docs

BarbourSmith:
The sprocket code is a direct port from the openSCAD project. I simplified it a little bit, but it's still not how I would have done things if I were starting from scratch. I won't be offended if you don't want it in here. I think it's a nice jumping off point at any rate.

I also added a note in .scale() that .scale(x,y,z) is possible because I noticed it was missing.

I tried to generate the docs this time around, we'll see if it worked 🤞. Let me know if I did it wrong. 

I made these changes on a mac computer starting from a fresh clone without hitting a single issue which was really great 😀


BarbourSmith:
Looking at the diffs I'm thinking that interactions_between_geometry/interactions_between_geometry.observed.md is not supposed to be checked in, is it? Am I supposed to rename it and save it?

pentacular:
> Looking at the diffs I'm thinking that interactions_between_geometry/interactions_between_geometry.observed.md is not supposed to be checked in, is it? Am I supposed to rename it and save it?

Yeah :)

All of the observed stuff shouldn't be in the PR.
Also the .html is an intermediate so we can drop that as well.

npm run doc

should output a list of commands to inspect and fix expectation failures at the end.

e.g.

cp "...observed..." to "..."
git add ...

If you run those, and push it should fix the PR and the CI should then pass.

pentacular:
I've been using CamelCase for functions and camelCase for variables.

It might be useful to follow that convention, but it's not critical.

nb/projects/belt_sprockets/belt_sprockets.nb
```diff
@@ -0,0 +1,683 @@
...
+
+  const additional_tooth_width = 0.2; //mm
+
+  //	If you need more tooth depth than this provides, adjust the following constant. However, this will cause the shape of the tooth to change.
+
+  const additional_tooth_depth = 0; //mm
+
+  // Functions
+
+  const tooth_spaceing_curvefit = (b, c, d) => {
```
pentacular, line 34: How about ToothSpacingCurvefit ?

nb/projects/belt_sprockets/belt_sprockets.nb
```diff
@@ -0,0 +1,683 @@
...
+    return ((c * Math.pow(teeth, d)) / (b + Math.pow(teeth, d))) * teeth;
+  };
+
+  const tooth_spacing = (tooth_pitch, pitch_line_offset) => {
+    return 2 * ((teeth * tooth_pitch) / (3.14159265 * 2) - pitch_line_offset);
```
pentacular, line 39: We don't need the return here

... => 2 * ...;

would be sufficient

nb/projects/belt_sprockets/belt_sprockets.nb
```diff
@@ -0,0 +1,683 @@
...
+      (tooth_width + additional_tooth_width) / tooth_width;
+    const tooth_depth_scale =
+      (tooth_depth + additional_tooth_depth) / tooth_depth;
+
+    const center = Arc(pulley_OD, pulley_OD);
+
+    //teeth - cut out of shaft
+
+    var toothProfile = [];
+    if (profile == 1) {
```
pentacular, line 576: I can't figure out where this variable is coming from :)

But I'd suggest just passing the profile in directly rather than having a lookup table.

nb/projects/belt_sprockets/belt_sprockets.nb
```diff
@@ -0,0 +1,683 @@
...
+    }
+    if (profile == 13) {
+      toothProfile = GT2_3mm;
+    }
+    if (profile == 14) {
+      toothProfile = GT2_5mm;
+    }
+
+    //Convert path to a face
+    toothProfile = Shape.fromOpenPath(toothProfile).close().fill();
```
pentacular, line 620: I'd consider moving this to the profile definitions, then you can just pass toothProfile in directly, and we can have a view of each profile in the notebook

nb/projects/belt_sprockets/belt_sprockets.nb
```diff
@@ -0,0 +1,683 @@
...
+  const HTD_5mm_pulley_dia = tooth_spacing(5, 0.5715);
+  const HTD_8mm_pulley_dia = tooth_spacing(8, 0.6858);
+  const GT2_2mm_pulley_dia = tooth_spacing(2, 0.254);
+  const GT2_3mm_pulley_dia = tooth_spacing(3, 0.381);
+  const GT2_5mm_pulley_dia = tooth_spacing(5, 0.5715);
+
+  // Tooth profile modules
+
+  // Tooth profile consts
+
```
pentacular, line 62: How about

const ToothProfile(...points) => Shape.fromOpenPath(points).close().fill();

nb/projects/belt_sprockets/belt_sprockets.nb
```diff
@@ -0,0 +1,683 @@
...
+    [0.370452, 0.48514, 0],
+    [0.39811, 0.458752, 0],
+    [0.416983, 0.424392, 0],
+    [0.540808, 0.083609, 0],
+    [0.559752, 0.049424, 0],
+    [0.587516, 0.023037, 0],
+    [0.621841, 0.006033, 0],
+    [0.660421, 0, 0],
+    [0.660421, -0.5, 0],
+  ];
```
pentacular, line 86: Then

const MXL = ToothProfile( [-0.660421, -0.5, 0], [-0.660421, 0, 0], ...).gridView().md('MXL tooth profile');

nb/projects/belt_sprockets/belt_sprockets.nb
```diff
@@ -0,0 +1,683 @@
...
+    return ((c * Math.pow(teeth, d)) / (b + Math.pow(teeth, d))) * teeth;
+  };
+
+  const tooth_spacing = (tooth_pitch, pitch_line_offset) => {
+    return 2 * ((teeth * tooth_pitch) / (3.14159265 * 2) - pitch_line_offset);
+  };
+
+  // The following set the pulley diameter for a given number of teeth
+
+  const MXL_pulley_dia = tooth_spacing(2.032, 0.254);
```
pentacular, line 44: Maybe the tooth spacing belongs with the tooth profile?

const MXL = { profile: ..., diameter: ToothSpacing(...) }; ?

Then we can pass those in as a unit.

nb/projects/belt_sprockets/belt_sprockets.nb
```diff
@@ -0,0 +1,683 @@
...
+// https://github.com/prusajr/PrusaMendel by Josef Prusa
+// http://www.thingiverse.com/thing:3104 by GilesBathgate
+// http://www.thingiverse.com/thing:2079 by nophead
+
+// dxf tooth data from http://oem.cadregister.com/asp/PPOW_Entry.asp?company=915217&elementID=07807803/METRIC/URETH/WV0025/F
+// pulley diameter checked and modeled from data at http://www.sdp-si.com/D265/HTML/D265T016.html
+
+// tune-able constants
+
+const makePulley = (profile, teeth) => {
```
pentacular, line 17: I didn't realize everything was embedded in this when I first looked at the code.

I think I'd try turning things a bit inside out.

const fooTooth = { profile: ToothProfile(...).view(), spacing: ... };

const FooPulley = teeth => Pulley(fooTooth, teeth).view();

etc.

Which would pull the resources out the top level and make them viewable in the notebook.

nb/projects/belt_sprockets/belt_sprockets.nb
pentacular: Ah, I just fixed up Polygon() to handle this, but it isn't merged yet.

nb/projects/belt_sprockets/belt_sprockets.nb
BarbourSmith: Yeah this makes a ton of sense. Having those things be separate is asking for trouble

nb/projects/belt_sprockets/belt_sprockets.nb
BarbourSmith: I've been doing toothSpacingCurvefit, when do you do a capital first letter and when do you do lowercase?

BarbourSmith: Ah, as usual you already answered this :p "I've been using CamelCase for functions and camelCase for variables."

nb/projects/belt_sprockets/belt_sprockets.nb
BarbourSmith: I feel pretty strongly this is the right thing to do in some way, but I have the issue that the diameter relies on the number of teeth which wasn't clear because they were using a global variable for the number of teeth:

`const ToothSpacing = (tooth_pitch, pitch_line_offset) => 2 * ((teeth * tooth_pitch) / (3.14159265 * 2) - pitch_line_offset);
`

Would it be poor form to have diameter be a function(teeth)? Like {......, diameter: (teeth) => {ToothSpacing(teeth, 2.032, 0.254)}}   ?

That way the magic numbers of 2.032 and 0.254 are encapsulated in the object, but it can still be evaluated at runtime to determine the diameter?

BarbourSmith:
I've cleaned things up somewhat and un-inside-outed it, but I am very much still open to suggestions on how to make things better 😀

BarbourSmith:
I'm failing tests, but it looks like its something in .dxf. The last pull request passed so it seems to be my fault, but I'm not sure how I broke that 😛 . Any thoughts?

pentacular:
> I'm failing tests, but it looks like its something in .dxf. The last pull request passed so it seems to be my fault, but I'm not sure how I broke that stuck_out_tongue . Any thoughts?

Not your fault -- a library must have shifted under us. :)

I'll sort it out.

nb/projects/belt_sprockets/belt_sprockets.nb
pentacular: I think that's a very nice solution. :)

nb/projects/belt_sprockets/belt_sprockets.nb
```diff
@@ -0,0 +1,741 @@
...
+
+const ToothSpacing = (teeth, tooth_pitch, pitch_line_offset) =>
+  2 * ((teeth * tooth_pitch) / (3.14159265 * 2) - pitch_line_offset);
+
+const ToothProfileToFace = (...points) =>
+  Shape.fromOpenPath(points).close().fill();
+
+const MXL = {
+  profile: ToothProfileToFace(
+    ...[
```
pentacular, line 42: ...[foo] is the same as foo, so you can just remove this outer layer.

nb/projects/belt_sprockets/belt_sprockets.nb
```diff
@@ -0,0 +1,741 @@
...
+      [0.39811, 0.458752, 0],
+      [0.416983, 0.424392, 0],
+      [0.540808, 0.083609, 0],
+      [0.559752, 0.049424, 0],
+      [0.587516, 0.023037, 0],
+      [0.621841, 0.006033, 0],
+      [0.660421, 0, 0],
+      [0.660421, -0.5, 0],
+    ]
+  ),
```
pentacular, line 66: The only think I'd suggest is that it would be nice to be able to see what the profiles look like, so I'd use Polygon(...).gridView().md('MXL profile') instead of Polygon(...)

nb/projects/belt_sprockets/belt_sprockets.nb
```diff
@@ -0,0 +1,741 @@
...
+};
+
+const ToothSpacing = (teeth, tooth_pitch, pitch_line_offset) =>
+  2 * ((teeth * tooth_pitch) / (3.14159265 * 2) - pitch_line_offset);
+
+const ToothProfileToFace = (...points) =>
+  Shape.fromOpenPath(points).close().fill();
+
+const MXL = {
+  profile: ToothProfileToFace(
```
pentacular, line 41: The profile shape won't get cached while it's in here, which I think isn't a problem.

But if you wanted it to be cached you could restructure it as

const xmlProfile = ToothProfileToFace(...);

const xml = { profile: xmlProfile, ... };

nb/projects/belt_sprockets/belt_sprockets.nb
```diff
@@ -0,0 +1,741 @@
...
+
+const ToothSpaceingCurvefit = (teeth, b, c, d) => {
+  return ((c * Math.pow(teeth, d)) / (b + Math.pow(teeth, d))) * teeth;
+};
+
+const ToothSpacing = (teeth, tooth_pitch, pitch_line_offset) =>
+  2 * ((teeth * tooth_pitch) / (3.14159265 * 2) - pitch_line_offset);
+
+const ToothProfileToFace = (...points) =>
+  Shape.fromOpenPath(points).close().fill();
```
pentacular, line 38: Polygon(...points).fill() would be nicer -- Shape.fromOpenPath is a bit low level -- but we can fix that later.

nb/projects/belt_sprockets/belt_sprockets.nb
```diff
@@ -0,0 +1,741 @@
...
+# Supported belt profiles:`;
+
+
+
+
+
+
+md`
+MXL`;
+Pulley(MXL, 20).view();
```
pentacular, line 679: If we're computing these, we might as well export definitions.

How about

export const xml20Pulley = Pulley(XML, 20).view().md('MXL'); ?

Or md('XML').view(); if you want the label above.

nb/projects/belt_sprockets/belt_sprockets.nb
```diff
@@ -0,0 +1,741 @@
...
+      [1.975908, 0, 0],
+      [1.975908, -0.75, 0],
+    ]
+  ),
+  diameter: (teeth) => ToothSpacing(teeth, 5, 0.5715),
+  toothDepth: 1.969,
+  toothWidth: 3.952,
+};
+
+const Pulley = (profile, teeth) => {
```
pentacular, line 636: This looks like the main entry point.

How about exporting it?

export const Pulley = ...

nb/projects/belt_sprockets/belt_sprockets.nb
```diff
@@ -0,0 +1,741 @@
...
+  return ((c * Math.pow(teeth, d)) / (b + Math.pow(teeth, d))) * teeth;
+};
+
+const ToothSpacing = (teeth, tooth_pitch, pitch_line_offset) =>
+  2 * ((teeth * tooth_pitch) / (3.14159265 * 2) - pitch_line_offset);
+
+const ToothProfileToFace = (...points) =>
+  Shape.fromOpenPath(points).close().fill();
+
+const MXL = {
```
pentacular, line 40: How about exporting the profile definitions?

export const XML = ...;

pentacular:
Looking very nice. :)

nb/projects/belt_sprockets/belt_sprockets.nb
BarbourSmith: This isn't supported by Polygon right now, right?

nb/projects/belt_sprockets/belt_sprockets.nb
BarbourSmith: Good catch :)

nb/projects/belt_sprockets/belt_sprockets.nb
BarbourSmith: I feel like its a small enough bit of memory that it's not a big concern, but its good to keep in mind

nb/projects/belt_sprockets/belt_sprockets.nb
BarbourSmith: 👍 👍

nb/projects/belt_sprockets/belt_sprockets.nb
BarbourSmith: Are they really useful? What are the odds that someone needs exactly 20 teeth?

BarbourSmith: The change in how the labels are generated is nice and much cleaner

nb/projects/belt_sprockets/belt_sprockets.nb
BarbourSmith: Yeah, that's pretty important :p

nb/projects/belt_sprockets/belt_sprockets.nb
BarbourSmith: Done

nb/projects/belt_sprockets/belt_sprockets.nb
pentacular: Should be if you sync and merge master

nb/projects/belt_sprockets/belt_sprockets.nb
pentacular: sg

nb/projects/belt_sprockets/belt_sprockets.nb
pentacular: Hmm, good point. :)

BarbourSmith:
I think I am done with changes to this for now so feel free to merge it if you think it's ready. Thanks for all the advice.

PR #1041, opened 2021-11-10
Support inverse transforms for items and segments, and allow use of edges as selectors with at and to

pentacular:
Shape.at now applies the ops sequentially at each selected orientation then restores to the original orientation.

Shape.to now duplicates the shape at each selected orientation.


geometry/tagged/getInverseMatrices.js
```diff
@@ -0,0 +1,40 @@
...
+  switch (geometry.type) {
+    case 'item': {
+      // These maintain an invertible matrix.
+      const global = geometry.matrix;
+      const local = invertTransform(global);
+      return { global, local };
+    }
+    case 'segments': {
+      // This is a bit trickier.
+      // We transform the matrices such that the first segment starts at [0, 0, 0], and extends to [length, 0, 0].
```
BarbourSmith, line 19: 👍

nb/projects/plotter/mount.nb
```diff
@@ -32,7 +32,7 @@ const upsideClip = Box(30 + 4)
       .z(10.5, 0)
       .x(23)
       .hull()
-      .op((f) => s.cut(f.extrudeAlong(normal(), -15)).and(f.ex(0.4)))
+      .op((f) => s.cut(f.extrudeAlong(normal(), -15)).and(f.extrudeAlong(normal(), 0.4)))
```
BarbourSmith, line 35: This is neat! I remember talking about the idea, but I didn't realize it was already working.

BarbourSmith (approved):
I'm a little hazy tonight, but to the best of my understanding this looks fantastic.

PR #1042, opened 2021-11-14
New main ui app

pentacular:


api/shape/table.js
```diff
@@ -0,0 +1,26 @@
+import { emit, generateUniqueId } from '@jsxcad/sys';
+
+import Shape from './Shape.js';
+import hashSum from 'hash-sum';
+import { md } from './md.js';
+
+export const table =
```
BarbourSmith, line 7: This makes sense

api/shape/view.js
```diff
@@ -9,11 +9,29 @@ import {
 import { Shape } from './Shape.js';
 import { ensurePages } from './Page.js';
 
+const byType = (args, defaultOptions) => {
```
BarbourSmith, line 12: This makes it so that the ({}, ...) object is optional? That's a neat idea if I'm understanding it right.

BarbourSmith (approved):
Looks good to me 👍 👍

api/shape/view.js
pentacular: It means that the order of the arguments doesn't matter -- they get assigned by type.

It expects one or zero Object, Function, and other arguments, in any order.

This is mostly to avoid needing to add a viewId to every view, and to deal with the current undefined arguments.

e.g., x.view(undefined, { skin: false }) becomes x.view({ skin: false }), and x.view(rx(1/2)) stays x.view(rz(1/2)).

I'm not sure this is a great decision, but I guess we'll see how it pans out.
I suspect viewIds might end up getting added in every case, and if so we can bring that back to being positional.

I'll update stl, etc, to use the same trick, for parity.

PR #1057, opened 2021-11-23
Add basic anchors, shadows, move display outline and wireframe to threejs

pentacular:


BarbourSmith (approved):
This pull request looks great to me, I'll give it a try and see if I have any thoughts on the user experience when it is merged

PR #1059, opened 2021-11-27
Support for orient rewrite

pentacular:


api/shape/Box.js
```diff
@@ -31,7 +31,7 @@ Shape.registerReifier('Box', (geometry) => {
   ]);
   const b = a.fill();
   const c = b.ex(top, bottom);
-  const d = c.orient({
+  const d = c.orient2({
```
BarbourSmith, line 34: Is the "2" here meant to read like "To"?

api/shape/Shape.js
```diff
@@ -270,7 +270,46 @@ Shape.toCoordinate = (x) => {
...
     return x;
+  } else {
+    return [x, y, z];
+  }
+};
+
+Shape.toCoordinates = (...args) => {
+  const coordinates = [];
+  while (args.length > 0) {
+    const x = args.shift();
```
BarbourSmith, line 281: I didn't know about .shift() and had to look it up. That's neat 👍

api/shape/orient.js
```diff
@@ -1,25 +1,66 @@
...
+      if (Math.abs(u[Z]) === 1) {
+        z[X] += 0.0001;
+      } else {
+        z[Z] += 0.0001;
+      }
+      z = normalize(z);
+      x = cross(u, z);
+    }
+    x = normalize(x);
+    let y = cross(z, x);
```
BarbourSmith, line 41: Nice work doing all this without any cos or sin

api/shape/z.js
```diff
@@ -4,6 +4,6 @@ import move from './move.js';
 export const z =
   (...z) =>
   (shape) =>
-    Shape.Group(...shape.toFlatValues(z).map((z) => move(0, 0, z)(shape)));
+    Shape.Group(...shape.toFlatValues(z).map((z) => move([0, 0, z])(shape)));
```
BarbourSmith, line 7: Are we moving away from (x,y,z) and towards ([x,y,z]) being the standard?

compiler/rewrite.js
```diff
@@ -1,17 +1,151 @@
...
+        );
+        if (!calleeObject) {
+          return;
+        }
+        const args = calleeObject.get('arguments');
+        const nthArg = args.value[nth];
+        if (!nthArg) {
+          return;
+        }
+        // Ok, we should have extracted an x of Group(x).view(viewId).
```
BarbourSmith, line 90: This is super neat

BarbourSmith (approved):
Wow this one is a lot. It look good to me 👍 👍

api/shape/Box.js
pentacular: Actually, it's just because I couldn't deprecate it quite yet. :)

I'll remove it in a while

api/shape/z.js
pentacular: Not exactly, rather moving toward them both coexisting.

But this form is a bit simpler to process and less ambiguous.

PR #1060, opened 2021-11-29
Retain matrix for triangles, and stop depending on gitcdn

pentacular:


api/shape/Shape.js
```diff
@@ -261,24 +261,46 @@ Shape.toNestedValues = (to, from) => {
   }
 };
 
-Shape.toCoordinate = (x = 0, y = 0, z = 0) => {
+Shape.toCoordinate = (shape, x = 0, y = 0, z = 0) => {
+  if (x instanceof Function) {
```
BarbourSmith, line 265: I understand

```
if (typeof x === 'string') {
    x = shape.get(x);
  }
```

But how does instanceof Function work in code?

api/shape/Shape.js
```diff
@@ -261,24 +261,46 @@ Shape.toNestedValues = (to, from) => {
...
       return g.points[0];
+    } else {
+      throw Error(`Unexpected coordinate value: ${x}`);
     }
   } else if (x instanceof Array) {
     return x;
-  } else {
+  } else if (typeof x === 'number') {
+    if (typeof y !== 'number') {
+      throw Error(`Unexpected coordinate value: ${y}`);
```
BarbourSmith, line 283: 👍

BarbourSmith (approved):
The 185 files changed was intimidating at first, but looking through it it mostly makes sense

PR #1061, opened 2021-11-29
Re-enable search and link clicking in the editor and allow viewing of purely void geometry

pentacular:


BarbourSmith (approved):
Looks good to me 👍

PR #1063, opened 2021-12-01
Add simple jogging and update default view to include grid

pentacular:


api/shape/view.js
```diff
@@ -58,7 +58,7 @@ export const baseView =
       height = 256,
       position = [100, -100, 100],
       withAxes = false,
-      withGrid = false,
+      withGrid = true,
```
BarbourSmith, line 61: :)

ui/threejs/moveToFit.js
```diff
@@ -16,7 +16,7 @@ export const moveToFit = ({
   scene,
   fitOffset = 1.2,
   withGrid = false,
-  gridLayer = SKETCH_LAYER,
+  gridLayer = GEOMETRY_LAYER,
```
BarbourSmith, line 19: What does it mean that the grid is on the geometry layer now?

BarbourSmith (approved):
Looks great to me, I'm excited to try them 👍

ui/threejs/moveToFit.js
pentacular: It means that it won't be drawn over the top of geometry.

PR #1064, opened 2021-12-02
AnchorControls finds best fit between camera and world axes

pentacular:


BarbourSmith (approved):
Clear concise and seems like it does what it says it does 👍 👍

PR #1065, opened 2021-12-03
Editable Assembly and allow cut, paste, and delete of shapes.

pentacular:


ui/threejs/AnchorControls.js
```diff
@@ -233,134 +186,144 @@ class AnchorControls extends EventDispatcher {
...
-          break;
-
-        // at
-        case 'd':
-          _at.position.addScaledVector(_xAxis, _step);
-          if (_lockUp) {
-            _up.position.addScaledVector(_xAxis, _step);
-          }
-          if (_lockTo) {
+      if (event.getModifierState('Control')) {
```
BarbourSmith, line 189: Mac users might expect this to be apple+C...but I'm comfortable with that not working

BarbourSmith (approved):
No big questions on my end 😀

ui/threejs/AnchorControls.js
pentacular: I think apple has the "Copy" key, etc, but I might be confused.

But we can add those in.

PR #1069, opened 2021-12-10
Various strategies for cut, and profiling timers

pentacular:


api/shape/startTimer.js
```diff
@@ -0,0 +1,10 @@
+import Shape from './Shape.js';
+
+export const starts = new Map();
+
+export const startTimer = (name) => (shape) => {
```
BarbourSmith, line 5: This is a nice feature to have generally

geometry/tagged/difference.js
```diff
@@ -1,17 +1,56 @@
...
-import { getGraphs } from './getGraphs.js';
 import { difference as graphDifference } from '../graph/difference.js';
 import { rewrite } from './visit.js';
 import { toConcreteGeometry } from './toConcreteGeometry.js';
 
-export const difference = (geometry, ...geometries) => {
-  geometries = geometries.map(toConcreteGeometry);
+export const difference = (geometry, options = {}, ...geometries) => {
+  if (
+    !['incremental', 'single', 'single_recursive', 'basic', undefined].includes(
```
BarbourSmith, line 15: So all of the modes are exposed now in the api which is nice for testing. Long term do you think all of them will be exposed or if there is a clear winner will just one be exposed in the end?

BarbourSmith (approved):
👍 👍 😀 

I'm going to have to update my local copy and play with these

api/shape/startTimer.js
pentacular: Yeah, I thought I had this already, but I couldn't find it. :)

PR #1071, opened 2021-12-14
Change default cut strategy to 'incremental' and fail on zero volume extrusions.

pentacular:


algorithm/cgal/extrudeSurfaceMesh.test.js
```diff
@@ -99,3 +99,25 @@ test('Extrude/Triangle', (t) => {
     isClosed: true,
   });
 });
+
+test('Zero volume extrusion fails', (t) => {
```
BarbourSmith, line 103: 👍 👍

BarbourSmith (approved):
Looks good to me, although I think I am not seeing most of the changes since the since the diffs for the cgal stuff aren't shown.

PR #1074, opened 2021-12-15
Add Assembly2 and some fixes

pentacular:


algorithm/cgal/cgal.cc
```diff
@@ -1737,146 +1737,62 @@ const double kExtrusionMinimumSquared = kExtrusionMinimum * kExtrusionMinimum;
...
-      SurfaceMeshSectionToPolygonSet(plane, *b, other);
-      set.difference(other);
-      return GeneralPolygonSetToSurfaceMesh(plane, set);
-    }
-  } else if (IsPlanarSurfaceMesh(plane, *b)) {
-    return a;
-  }
+// Use different iota to avoid 45 degree slides.
+const double kIotaX = 10e-5;
+const double kIotaY = 11e-5;
```
BarbourSmith, line 1742: 😀

api/shape/main.js
```diff
@@ -355,6 +355,7 @@ export { z } from './z.js';
 export { Alpha } from './Alpha.js';
 export { Arc } from './Arc.js';
 export { Assembly } from './Assembly.js';
+export { Assembly2 } from './Assembly2.js';
```
BarbourSmith, line 358: So it's actually called Assembly2 in the api? Like to test I should call it that way?

BarbourSmith (approved):
Exciting stuff :)

api/shape/main.js
pentacular: Yeah, I wanted to keep it an experimental interface for now, since it doesn't address the general problem -- just volumes for now.

PR #1078, opened 2021-12-18
Update Box and Arc to use extents. Add Arc[XYZ], e[xyz], etc

pentacular:


api/shape/extrudeAlong.js
```diff
@@ -33,8 +33,9 @@ export const extrudeAlong =
     return Shape.Group(...extrusions);
   };
 
-export const ex = (...extents) => extrudeAlong(Point(0, 0, 1), ...extents);
+export const e = (...extents) => extrudeAlong(normal(), ...extents);
```
BarbourSmith, line 36: I'm betting I write this one wrong a couple times before I get used to it 😛 

I think I'm going to make .ez() my new default tho instead of using .e, because like you said it's more precise

BarbourSmith (approved):
Looks good to me.

api/shape/extrudeAlong.js
pentacular: Yeah, normal inference works by fitting a plane to the points and then preferring a positive orientation on one of the axes, so it's a bit of a best effort.

pentacular: Ah, once we preserve the local frame of reference properly it should probably become the initial z axis which will be much more solid -- I guess it's a conflict between geometric normal (which is what normal() does now), and initial normal.

PR #1079, opened 2021-12-19
Configure ui and configure workers upon attaching

pentacular:


ui/app/App.js
```diff
@@ -386,6 +395,46 @@ class App extends React.Component {
...
+        const step = steps.shift();
+        if (object[step] === undefined) {
+          object[step] = {};
+        }
+        object = object[step];
+      }
+      return object;
+    };
+
+    this.Config.update = async () => {
```
BarbourSmith, line 414: Basically this change allows the worker configuration to be updated rather than compiled in?

ui/app/webworker.js
```diff
@@ -62,6 +63,7 @@ const agent = async ({ ask, message, type, tell }) => {
       }
       case 'sys/attach':
         self.id = id;
+        sys.setConfig(config);
```
BarbourSmith, line 66: Do I need to make these changes to my worker to keep it up to date?

BarbourSmith (approved):
Looks good to me!

ui/app/App.js
pentacular: Yeah, that's the idea. This bit handles persistent storage of the configuration, the actual communication is via sys/attach

ui/app/webworker.js
pentacular: Only if you want to send config to the workers -- it'll use the compiled in defaults otherwise.

BarbourSmith: Very nice.

PR #1082, opened 2021-12-30
Update clip

pentacular:


api/shape/Line.js
```diff
@@ -1,14 +1,10 @@
-import Path from './Path.js';
```
BarbourSmith, line 1: Are paths still a thing?

api/shape/seq.js
```diff
@@ -2,22 +2,40 @@ import Shape from './Shape.js';
...
 
+const maybeApply = (value, shape) => {
+  if (value instanceof Function) {
+    return value(shape);
+  } else {
+    return value;
+  }
+};
+
+// This is getting a bit excessively magical.
```
BarbourSmith, line 13: 😛

BarbourSmith (approved):
Looks great to me. I'll update immediately and test things today.

PR #1083, opened 2021-12-31
Fix layout bounds and add svg generation regression case

pentacular:


BarbourSmith (approved):
This is a nice compact PR compared to the last couple 😀 

I'll update immediately.

PR #1084, opened 2021-12-31
Fix non-intersecting clip, fix layout bounding box, multidimensional seq, points refactoring

pentacular:


algorithm/cgal/cgal.cc
```diff
@@ -1705,27 +1705,57 @@ void convertArrangementToPolygonsWithHoles(
...
       do {
-        if (edge->source()->point() == edge->target()->point()) {
+        const Point_2& point = edge->source()->point();
+        if (point == edge->target()->point()) {
           // Skip zero length edges.
           continue;
         }
-        polygon_hole.push_back(edge->source()->point());
+        if (polygon_hole.size() >= 2 &&
+            CGAL::collinear(polygon_hole.end()[-2], polygon_hole.end()[-1],
```
BarbourSmith, line 1743: This is the part that removed the extra points?

BarbourSmith (approved):
Looks great to me, I'm excited to play with the new sequence options.

algorithm/cgal/cgal.cc
pentacular: Part of it -- it checks sequences of three vertices for colinearity, and replaces or removes the central one if it finds it.

PR #1088, opened 2022-01-06
Support separate() for surfaces

pentacular:


algorithm/cgal/separateSurfaceMesh.js
```diff
@@ -2,18 +2,21 @@ import { getCgal } from './getCgal.js';
 
 export const separateSurfaceMesh = (
   mesh,
-  keepVolumes = true,
-  keepCavitiesInVolumes = true,
-  keepCavitiesAsVolumes = false
+  keepShapes = true,
```
BarbourSmith, line 5: That's interesting that keeping only the holes is an option

BarbourSmith (approved):
Looks great to me!

algorithm/cgal/separateSurfaceMesh.js
pentacular: Yeah, it means that you can extract the cavities in a shape as shapes.

e.g., given Box(10).cut(Arc(3)) you could get back the Arc(3)

PR #1089, opened 2022-01-07
Parallel cut algorithm

pentacular:


algorithm/cgal/cgal.cc
```diff
@@ -85,6 +87,10 @@
 #include <boost/range/adaptor/reversed.hpp>
 #include <queue>
 
+#ifdef CUSTOM_HAS_THREADS
+#include <thread>
```
BarbourSmith, line 91: Its cool that this thread stuff can stay in for when it becomes stable

algorithm/cgal/cgal.cc
```diff
@@ -2119,6 +2079,198 @@ int ClipSurfaceMeshes(size_t target_count, emscripten::val getTargetMesh,
   return STATUS_OK;
 }
 
+int ParallelCutSurfaceMeshes(size_t target_count, emscripten::val getTargetMesh,
```
BarbourSmith, line 2082: Right now this one isn't used right? But we can switch to it once cgal becomes thread safe?

algorithm/cgal/cutSurfaceMeshes.js
```diff
@@ -9,7 +9,7 @@ export const cutSurfaceMeshes = (targetMeshes, targetSegments, sources) => {
     const cutMeshes = [];
     const cutSegments = [];
     let nthSegments;
-    const status = getCgal().CutSurfaceMeshes(
+    const status = getCgal().ParallelCutSurfaceMeshes(
```
BarbourSmith, line 12: Or no it is used. Is this parallel done using workers? If so where is the worker called from?

BarbourSmith (approved):
lgtm 👍 👍

algorithm/cgal/cgal.cc
pentacular: Actually, this is the one in use at the moment -- it's just compiled with threading off.

We should see if there's a performance difference between the two algorithms.

algorithm/cgal/cutSurfaceMeshes.js
pentacular: It's a parallel algorithm, but in this case we're running the parallel components sequentially since threading is disabled.

PR #1090, opened 2022-01-09
Add simplify shape, mesh, isotropic remeshing with area selection

pentacular:


algorithm/cgal/isotropicRemeshingOfSurfaceMesh.js
```diff
@@ -0,0 +1,26 @@
+import { getCgal } from './getCgal.js';
+import { toCgalTransformFromJsTransform } from './transform.js';
+
+export const isotropicRemeshingOfSurfaceMesh = (
+  mesh,
+  matrix,
+  { iterations = 1, relaxationSteps = 1, targetEdgeLength = 1.0 },
```
BarbourSmith, line 7: What are relaxation steps?

algorithm/cgal/smoothSurfaceMesh.js
```diff
@@ -0,0 +1,25 @@
+import { getCgal } from './getCgal.js';
+import { toCgalTransformFromJsTransform } from './transform.js';
+
+export const smoothSurfaceMesh = (
+  mesh,
+  matrix,
+  { iterations = 1, safe = true },
```
BarbourSmith, line 7: What does safe do here?

geometry/graph/smooth.js
```diff
@@ -1,28 +1,97 @@
...
+            options,
+            selectionGraphs.map(({ graph, matrix }) => ({
+              mesh: toSurfaceMesh(graph),
+              matrix,
+            }))
+          )
+        )
+      );
+    }
+    case 'SmoothMesh': {
```
BarbourSmith, line 48: What are the differences between these?

nb/projects/lego_sorter/lego_sorter.md
```diff
@@ -0,0 +1,65 @@
...
+
+```JavaScript
+bentWall.and(ring, base, ring.z(52)).scale(1.5).view().stl('tall_sorter');
+```
+
+![Image](lego_sorter.md.6.png)
+
+![Image](lego_sorter.md.7.png)
+
+[tall_sorter_0.stl](lego_sorter.tall_sorter_0.stl)
```
BarbourSmith, line 65: This looks cool! Does it sort by size when shaken?

nb/regression/shape/shape.nb
```diff
@@ -192,3 +192,11 @@ Arc(2)
   .view();
 
 Triangle(14).svg('triangle');
+
+Box(10)
+  .ez(50, -50)
+  .rx(0, 1 / 4)
+  .fuse().view(0)
+  .smooth({ method: 'IsotropicRemeshing', iterations: 20, relaxationSteps: 10, targetEdgeLength: 0.5, }, Box(20, 20, 20))
```
BarbourSmith, line 200: So the Box(20, 20, 20) argument to .smooth() is doing the selection? I like that a lot.

I was picturing something that worked like a.select(Box(20,20,20)).smooth() when you described it (I got your message but didn't have service to reply), but that has all sorts of tricky issues about how it would actually work.

BarbourSmith (approved):
Looks good to me, lots of questions but no comments 😁 

After this merges if I want to play with it is https://jsxcad.js.org/alpha/ still the place?

nb/regression/shape/shape.nb
pentacular: Yes, some kind of sticky section could be done, but I think as long as we can make the following work, we should be fine.

a.and(Box(20, 20, 20).opaque(0.3).as('selection')).smooth(get('selection')).view()

BarbourSmith: Yeah, that looks nice and should work now, right?

PR #1091, opened 2022-01-11
Update remesh, smooth, simplify.

pentacular:


es6/jsxcad-ui-threejs.js
```diff
@@ -44927,6 +44930,7 @@ const orbitDisplay = async (
     renderer,
     scene,
     trackballControls,
+    updateFit,
```
BarbourSmith, line 44933: Wow that was fast I didn't expect this change to be in here already

BarbourSmith (approved):
Looks good to me! Thanks for the addition 🙂

PR #1097, opened 2022-01-15
Update hashing, cache cut, cache thumbnails.

pentacular:


BarbourSmith (approved):
I can't say I fully understand every change in this one, but I am excited to try it and can hopefully do some testing today.

PR #1099, opened 2022-01-18
Add caching for join and clip, restructure disjoint. Various refactorings.

pentacular:


api/shape/Box.js
```diff
@@ -11,6 +11,49 @@ const X = 0;
 const Y = 1;
 const Z = 2;
 
+let fundamentalShapes;
+
+const fs = () => {
+  if (fundamentalShapes === undefined) {
+    fundamentalShapes = {
+      tlfBox: Point(),
```
BarbourSmith, line 19: Not complaining, just curious. Why are these different for the different faces? Is it because rotation is not exact at the moment? If the rounding thing works out and rotation can be expressed exactly would these become one?

BarbourSmith (approved):
lgtm 👍 👍

api/shape/Box.js
pentacular: That's part of it, the other part is so that the inverse transforms lead back to this orientation.

We could also have achieved that by rotating, and then applying the rotation to the mesh directly, but this seemed simpler.

BarbourSmith: That's a really nice way to keep track of what the original orientation was!

PR #1101, opened 2022-01-19
Phase out approximate matrices, and add more flexible double to rational translation.

pentacular:


api/shape/orient.js
```diff
@@ -57,6 +57,8 @@ export const orient =
       0,
       1,
     ];
+    // FIX: Move this to CGAL.
+    lookAt.blessed = true;
```
BarbourSmith, line 61: Blessed won't be a thing long term, right? It's only to ease the transition?

BarbourSmith (approved):
Super impressive as always. No big thoughts from me :)

PR #1102, opened 2022-01-20
Apply the direction operation to individual leaves in extrudeAlong, and have computeNormal return the last leaf's normal.

pentacular:


nb/regression/shapes/shapes.md
```diff
@@ -394,3 +394,11 @@ Box(12)
...
 
 ![Image](shapes.md.33.png)
+
+```JavaScript
+Group(Box(1), Box(1).rx(1 / 4))
+  .e(2)
+  .view();
+```
+
+![Image](shapes.md.34.png)
```
BarbourSmith, line 404: That's the behavior I would expect here

BarbourSmith (approved):
👍👍👌

PR #1104, opened 2022-01-24
Toolpath, segments, etc.

pentacular:


algorithm/cgal/fitPlaneToPoints.js
```diff
@@ -11,8 +11,13 @@ export const fitPlaneToPoints = (points) => {
     const plane = [0, 0, 1, 0];
     c.FitPlaneToPoints(
       (triples) => {
+        let count = 0;
         for (const [x, y, z] of points) {
           c.addDoubleTriple(triples, x, y, z);
+          if (count++ > 1000) {
+            // We're blowing the stack when this is too high.
```
BarbourSmith, line 18: 😛 

Is it really as low as 1000?

geometry/tagged/computeToolpath.js
```diff
@@ -0,0 +1,201 @@
...
+import { outline } from './outline.js';
+import { section } from './section.js';
+import { taggedToolpath } from './taggedToolpath.js';
+import { toConcreteGeometry } from './toConcreteGeometry.js';
+
+const X = 0;
+const Y = 1;
+const Z = 2;
+
+export const computeToolpath = (geometry, toolDiameter = 1, jumpHeight = 1) => {
```
BarbourSmith, line 17: This is impressively compact for everything that it's doing

BarbourSmith (approved):
It all looks good to me 😀 

Some of the .pngs in plotter aren't showing the same thing that they were before, but it doesn't look wrong. Are they different because toolpaths aren't in soup so they don't show up in .png?

geometry/tagged/computeToolpath.js
pentacular: Yeah, but it's also doing things very inefficiently. :)

I suspect it'll get a bit larger.

PR #1105, opened 2022-01-29
Refactor arrangements and simplified polygon points in construction

pentacular:


algorithm/cgal/arrangePolygonsWithHoles.test.js
```diff
@@ -46,12 +46,12 @@ test('arrangePolygonsWithHoles', (t) => {
...
           ],
           exactPoints: [
-            ['-5214694305376363/18014398509481984', '-1/2', '0'],
-            ['1/4', '-6816258895479669/36028797018963968', '0'],
-            ['7110946780058677/9007199254740992', '-1/2', '0'],
-            ['7110946780058677/9007199254740992', '1/2', '0'],
-            ['1/4', '6816258895479669/36028797018963968', '0'],
-            ['-5214694305376363/18014398509481984', '1/2', '0'],
+            ['-11/38', '-1/2', '0'],
+            ['1/4', '-7/37', '0'],
```
BarbourSmith, line 50: Well that's nice 😀

algorithm/cgal/cgal.cc
```diff
@@ -3930,6 +3929,169 @@ void print_polygon_with_holes(
...
+                           emscripten::val& emit_polygon,
+                           emscripten::val& emit_point) {
+  bool emitted = false;
+  for (const P& polygon : polygons) {
+    const auto& outer = polygon.outer_boundary();
+    emit_polygon(false);
+    emitted = true;
+    for (auto edge = outer.edges_begin(); edge != outer.edges_end(); ++edge) {
+      if (edge->source() == edge->target()) {
+        // Skip zero length edges.
```
BarbourSmith, line 3943: What does it mean for an edge to have zero length? Wouldn't that by definition not be contributing any new information so it could be skipped? 

...Or is that what's happening here

algorithm/cgal/cgal.cc
```diff
@@ -3930,6 +3929,169 @@ void print_polygon_with_holes(
...
+                             sin(-angle) * ox + cos(-angle) * oy + cy));
+    }
+    plan.push_back(Point_2(tx, ty));
+  } else {
+    // COUNTER-CLOCKWISE
+    plan.push_back(Point_2(sx, sy));
+    if (target_angle < source_angle) {
+      target_angle += M_PI * 2;
+    }
+    const double step = M_PI / 32.0;
```
BarbourSmith, line 4008: Is there any value to this being proportional to the diameter? Probably not worth the extra complexity

algorithm/cgal/cutSurfaceMeshes.js
```diff
@@ -46,7 +46,7 @@ export const cutSurfaceMeshes = (targetMeshes, targetSegments, sources) => {
     );
 
     if (status === STATUS_ZERO_THICKNESS) {
-      throw new ErrorZeroThickness('Zero thickness produced by join');
+      throw new ErrorZeroThickness('Zero thickness produced by cut');
```
BarbourSmith, line 49: Nice 👍

geometry/graph/offset.test.js
```diff
@@ -26,78 +26,46 @@ test('offset', (t) => {
...
-      '-11/18 2/3 1/2\n' +
-      '-25/39 25/39 1/2\n' +
-      '-15/26 13/19 1/2\n' +
-      '-2/3 11/18 1/2\n' +
-      '-16/23 7/13 1/2\n' +
-      '-13/19 15/26 1/2\n' +
+    '20\n' +
+      '-41/64 -41/64 1/2\n' +
+      '-3152519739159347/4503599627370496 -1/2 1/2\n' +
+      '-385724968084973/562949953421312 -2598230554252209/4503599627370496 1/2\n' +
```
BarbourSmith, line 32: Why did these ones seemingly get more complex instead of less?

BarbourSmith (approved):
Looks great to me :)

algorithm/cgal/cgal.cc
pentacular: Yeah, they shouldn't exist, really.

algorithm/cgal/cgal.cc
pentacular: I was thinking about it, but let's see how it goes first.

geometry/graph/offset.test.js
pentacular: Hmm, that's a good question.

PR #1107, opened 2022-02-04
Update toolpath generation

pentacular:


api/shape/toolpath.js
```diff
@@ -6,21 +6,39 @@ import { computeToolpath } from '@jsxcad/geometry';
...
 
 export const toolpath =
-  ({ diameter = 1, jumpHeight = 1 } = {}) =>
+  ({
+    diameter = 1,
+    jumpHeight = 1,
+    stepCost = diameter * -2,
+    turnCost = -2,
+    neighborCost = -2,
+    stopCost = 30,
```
BarbourSmith, line 15: This is cool! So tuning these will change what it prioritizes?

BarbourSmith (approved):
👍 👍 👍

api/shape/toolpath.js
pentacular: Yeah, although I'm not sure how much practical flexibility there will be.

PR #1109, opened 2022-02-07
Add packing envelopes

pentacular:


algorithm/cgal/cgal.cc
```diff
@@ -4779,6 +4695,105 @@ void FromSurfaceMeshToPolygonsWithHoles(const Surface_mesh* input,
...
+      PS::simplify(ct, cid, cost, stop, true);
+    }
+
+    // And emit.
+    for (CT::Constraint_id cid : cids) {
+      emit_polygon(false);
+      Vertices_in_constraint_iterator it = ct.vertices_in_constraint_begin(cid);
+      Vertices_in_constraint_iterator end = ct.vertices_in_constraint_end(cid);
+      if (it != end) {
+        for (;;) {
```
BarbourSmith, line 4766: I had to look up how this works, it's neat. I didn't realize that ALL of the arguments to for() are optional

algorithm/cgal/insetOfPolygonWithHoles.test.js
```diff
@@ -16,128 +16,136 @@ test('insetOfPolygonWithHoles/bow', (t) => {
...
+        [0.109375, 0.048564189189189186, 0],
+        [-0.0894736842105263, 0.1631282959789057, 0],
       ],
       exactPoints: [
-        ['-17/190', '-826757/5324670', '0'],
-        ['9/52', '-4/999', '0'],
-        ['373/1924', '0', '0'],
-        ['9/52', '4/999', '0'],
-        ['-17/190', '826757/5324670', '0'],
+        ['-17/190', '-79189/485440', '0'],
```
BarbourSmith, line 90: What was the change to inset? Was it the addition of segments?...or was that already there?

nb/projects/plotter/plotter.stars_0.gcode
```diff
@@ -2,377 +2,270 @@
 G21
```
BarbourSmith: We might want to add a G90 or G91 to the front of these to indicate absolute vs incremental moves. G90 is absolute.

BarbourSmith (approved):
Looks great!

algorithm/cgal/cgal.cc
pentacular: It's not that they're optional, but rather that an empty statement is valid. :)

algorithm/cgal/insetOfPolygonWithHoles.test.js
pentacular: The additional of parameterized segments, I think.

nb/projects/plotter/plotter.stars_0.gcode
pentacular: Ah, good point.

PR #1110, opened 2022-02-08
Generate upper envelopes.

pentacular:


nb/projects/plotter/plotter.spiral_0.gcode
```diff
@@ -1,4 +1,5 @@
 
 G21
+G90
```
BarbourSmith, line 3: 🙂

BarbourSmith (approved):
I don't have any big thoughts on this one, but it looks good to me 👍 👍

PR #1114, opened 2022-03-01
Initial deform implementation

pentacular:


api/shape/view.js
```diff
@@ -146,14 +146,20 @@ Shape.registerMethod('frontView', frontView);
...
   (shape) => {
-    const { viewId, op, options } = byType(args, {
-      size: 512,
-      skin: true,
-      outline: true,
-      wireframe: false,
-      width: 1024,
-      height: 512,
-      position: [100, 0, 0],
+    const {
```
BarbourSmith, line 149: What do these changes to view() do?

BarbourSmith (approved):
Looks good to me :)

api/shape/view.js
pentacular: Ah, it's just cleaning up the argument understanding.

It means that view('foo', rz(1/2), { wireframe: true }) and view(rz(1/2), 'foo', { wireframe: true }) etc are all equivalent.

PR #1116, opened 2022-03-03
Import svg and collada via threejs geometry

pentacular:


api/threejs/readCollada.js
```diff
@@ -0,0 +1,16 @@
+import {
+  fromColladaToThreejs,
+  fromThreejsToGeometry,
+} from '@jsxcad/convert-threejs';
+
+import { Shape } from '@jsxcad/api-shape';
+import { read } from '@jsxcad/sys';
+
+export const readCollada = async (path, { src, invert = false } = {}) => {
```
BarbourSmith, line 9: It's pretty sweet how compact these become once the piping is in

convert/threejs/fromColladaToThreejs.test.js
```diff
@@ -0,0 +1,26 @@
...
+  const threejs = await fromColladaToThreejs(data);
+  const geometry = JSON.parse(
+    JSON.stringify(
+      prepareForSerialization(await fromThreejsToGeometry(threejs))
+    )
+  );
+  const expectedGeometry = JSON.parse(
+    readFileSync('duck_triangles.json', 'utf8')
+  );
+  t.deepEqual(geometry, expectedGeometry);
```
BarbourSmith, line 25: Sometimes I forget how good the testing setup is. I think that was a critical decision getting the testing in early and maintaining it. We used to have regressions all the time, but they haven't been a problem in a while.

BarbourSmith (approved):
Looks great to me.


Out of curiosity...If we use three for STL generation will the resulting .STL file be any different? Or is the conversion from cgal geometry to stl geometry 1:1 enough that they should be identical?

pentacular:
> 

You might find a different arrangement of triangles, but I wouldn't expect it to matter much -- STL is pretty trivial.

BarbourSmith:
That's kind of what I was guessing

PR #1117, opened 2022-03-12
Handle polymorphic disjoint in cgal

pentacular:


geometry/tagged/disjoint.js
```diff
@@ -1,10 +1,11 @@
 import { rewrite, visit } from './visit.js';
 
 import { cut } from './cut.js';
+import { disjoint as disjoinGraphs } from '../graph/disjoint.js';
 import { taggedGroup } from './taggedGroup.js';
 import { toConcreteGeometry } from './toConcreteGeometry.js';
 
-export const disjoint = (geometries) => {
+export const disjoint = (geometries, strategy = 'incremental') => {
```
BarbourSmith, line 8: Is incremental the new one?

BarbourSmith (approved):
No big thoughts on the inner workings other than that I am excited to try it. I think using it in practice will be the real test so I'll update tomorrow if the tests pass and try to break things :p

geometry/tagged/disjoint.js
pentacular: Yeah, but you can switch back if it has problems.

BarbourSmith: Great 👍, I just wanted to make sure I tested the right one 😁

PR #1118, opened 2022-03-15
New geometry mediation for disjoint

pentacular:


algorithm/cgal/cgal.cc
```diff
@@ -3840,6 +3846,287 @@ int DisjointSurfaceMeshesWithIncrementalMasking(int meshCount,
...
+        for (int nth = start + 1; nth < input_count; nth++) {
+          // TODO: Disjunction of planar by volume.
+          if (is_masked[nth] || types[nth] != GEOMETRY_POLYGONS_WITH_HOLES ||
+              planes[start] != planes[nth] ||
+              !CGAL::do_overlap(bounds_2[start], bounds_2[nth])) {
+            continue;
+          }
+          gps[start]->difference(*gps[nth]);
+        }
+        // TODO: Handle disjunction of surface by volume.
```
BarbourSmith, line 4076: Will cgal be able to do these? I don't remember seeing anything in there relating to interacting 2d and 3d representations...Although I guess the volume could always be sectioned

BarbourSmith (approved):
Looks good to me :)

BarbourSmith:
It looks like some of the .stls which changed ever so slightly are coming up different on the CI computer for some reason...which is weird and seems tricky

algorithm/cgal/cgal.cc
pentacular: Well, it's not really 2d vs 3d.

You can clip or cut open meshes with closed meshes.

But, yeah -- otherwise you can take a section, which we can convert to polygons-with-holes, which is probably the right approach here.

pentacular:
> 

Yeah, it's disturbing.

I downloaded the ci output and checked it -- it seems equivalent.

I'll disable the stl comparison in the short term.

PR #1127, opened 2022-05-01
cutopen, clipopen, and various improvements

pentacular:


algorithm/cgal/cgal.cc
```diff
@@ -3844,7 +3792,7 @@ int Cast(Geometry* geometry, const Transformation* reference) {
   return STATUS_OK;
 }
 
-int Clip(Geometry* geometry, int targets) {
+int Clip(Geometry* geometry, int targets, bool open) {
```
BarbourSmith, line 3795: Neat :)

algorithm/cgal/section.test.js
```diff
@@ -135,8 +135,8 @@ test('Coplanar section of polygon is the polygon.', (t) => {
     plane: [0, 0, 1, 0],
   };
   const sections = section(
-    [geometry],
-    [{ matrix: blessed([...identityMatrix]) }]
+    [geometry, { type: 'points', matrix: blessed([...identityMatrix]) }],
```
BarbourSmith, line 138: What does it mean for a matrix to be blessed?

api/shape/eachIn.js
```diff
@@ -0,0 +1,15 @@
+import Shape from './Shape.js';
+import { getLeafsIn } from '@jsxcad/geometry';
+
+export const eachIn =
```
BarbourSmith, line 4: 👍

api/shape/masking.js
```diff
@@ -0,0 +1,13 @@
+import Group from './Group.js';
+import Shape from './Shape.js';
+import { hasTypeMasked } from '@jsxcad/geometry';
+
+export const masking = (masked) => (shape) =>
```
BarbourSmith, line 5: I don't think masking is documented, but I'm not sure how it works.

BarbourSmith (approved):
This looks fantastic 😀

algorithm/cgal/section.test.js
pentacular: We throw errors on unblessed matrices to try to track down where they're getting produced.

At this point it can probably be retired.

api/shape/masking.js
pentacular: Yeah, I've been a bit slack on that.

a.masking(b) === b.mask(a)

PR #1128, opened 2022-05-02
Update to and at; selections for grow; add seam

pentacular:


algorithm/cgal/cgal.cc
```diff
@@ -5243,6 +5290,32 @@ int Outline(Geometry* geometry) {
   return STATUS_OK;
 }
 
+int Seam(Geometry* geometry, size_t count) {
```
BarbourSmith, line 5293: Is Seam already working? I thought the plan was to add that at some point in the future?

nb/regression/shapes/shapes.md
```diff
@@ -413,10 +405,71 @@ Orb(10)
...
+![Image](shapes.md.39.png)
+
+```JavaScript
+blue.fitTo(red.to(blue)).view();
+```
+
+![Image](shapes.md.40.png)
+
+```JavaScript
+red.at(blue, cut(Box([10, 20], 20))).view();
```
BarbourSmith, line 466: Nice 🙂

BarbourSmith (approved):
Other than the tests, looks good to me 👍 👍

algorithm/cgal/cgal.cc
pentacular: Yeah, I wanted to see how it worked -- I think it'll require a bit of refinement, but it was easy enough to do.

PR #1129, opened 2022-05-12
Some updates to the documentation

BarbourSmith:
This adds all the things that I know are missing. Curve will need to be added soon, and .at() needs some love.

There is something going on with my local install so I can't build the notebooks right now 😐, but I've taken a look at all of these in /alpha and they look good there.

### DIFF
--- a/nb/documentation/interactions_between_geometry/interactions_between_geometry.nb
+++ b/nb/documentation/interactions_between_geometry/interactions_between_geometry.nb
@@ -100,7 +100,7 @@ Hull(Arc(10), Box(5).move(12, 12, 0), Arc(20).x(40)).view();
 md`
 ---
 ### Mask
-Mask is used to create a void around something. This allows the shape to cut a larger hole than it's actual size to add play or for example around the threads of a bolt.
+Mask is used to create a mask around something. This allows the shape to cut a larger hole than it's actual size to add play or for example around the threads of a bolt.
 `;
 
 const shaft = Arc(10, 10, 40).color('grey');
--- a/nb/documentation/interactions_between_geometry/interactions_between_geometry.nb
+++ b/nb/documentation/interactions_between_geometry/interactions_between_geometry.nb
@@ -63,6 +63,21 @@ Arc(10, 10, 5).color('blue').cut(Box(5, 5, 5).move(3, 0, 2).color('red')).view()
 
 Arc(10, 10, 5).cutfrom(Box(5, 5, 5).move(3, 0, 2)).view().md(`The result is red since the red shape was cut from.`);
 
+
+md`
+---
+### Fit
+Fits two shapes together such that one cuts into the other. The result is similar to assembly.
+`;
+
+const axle = Arc(5, 5, 10).color('brown');
+const wheel = Arc(20, 20, 2).color('grey');
+
+wheel.fitTo(axle).view();
+axle.fit(wheel).view();
+
+wheel.fitTo(axle).pack().view(); //In both of these cases the axle cuts the wheel.
+
 md`
 ---
 ### Group
@@ -81,6 +96,15 @@ md`
 Performs the hull operation on the input shapes.`;
 
 Hull(Arc(10), Box(5).move(12, 12, 0), Arc(20).x(40)).view();
+
+md`
+---
+### Mask
+Mask is used to create a void around something. Mostly used for adding play.
+`;
+
+const shaft = Arc(10, 10, 40).color('grey');
+shaft.mask(shaft.grow(1)).view();
 
 md`
 ---
--- a/nb/documentation/interactions_between_geometry/interactions_between_geometry.nb
+++ b/nb/documentation/interactions_between_geometry/interactions_between_geometry.nb
@@ -100,11 +100,11 @@ Hull(Arc(10), Box(5).move(12, 12, 0), Arc(20).x(40)).view();
 md`
 ---
 ### Mask
-Mask is used to create a void around something. Mostly used for adding play.
+Mask is used to create a void around something. This allows the shape to cut a larger hole than it's actual size to add play or for example around the threads of a bolt.
 `;
 
 const shaft = Arc(10, 10, 40).color('grey');
-shaft.mask(shaft.grow(1)).view();
+shaft.mask(grow(1)).view();
 
 md`
 ---
--- a/nb/documentation/interactions_with_geometry/interactions_with_geometry.nb
+++ b/nb/documentation/interactions_with_geometry/interactions_with_geometry.nb
@@ -38,6 +38,18 @@ Arc(4)
   .view();
 
 
+
+md`
+---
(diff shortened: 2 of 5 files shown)


nb/documentation/interactions_with_geometry/interactions_with_geometry.nb
```diff
@@ -8,6 +8,27 @@ Geometry can be given a color`;
...
+const aGroup = Group(
+  Box(10, 200, 10),
+  Box(10, 200, 10).z(20),
+  Box(10, 200, 10).z(40)
+).view();
+
+Group(
+  ...aGroup.eachIn(
+    (s) => s.rz(Math.random()),
+    (b) => b
```
BarbourSmith, line 28: How does the second argument here work? I should add an example for that

pentacular: The first function should transform individual elements.
The second function should transform the elements as a group.

e.g., a.eachIn(color('blue'), Link)

nb/documentation/interactions_between_geometry/interactions_between_geometry.nb
```diff
@@ -82,6 +97,15 @@ Performs the hull operation on the input shapes.`;
 
 Hull(Arc(10), Box(5).move(12, 12, 0), Arc(20).x(40)).view();
 
+md`
+---
+### Mask
+Mask is used to create a void around something. Mostly used for adding play.
```
pentacular, line 103: Rather, a mask is used instead of the masked thing for cutting.

e.g., you might want to mask a bolt thread with a simple cylinder to cut a bolt hole rather than threads through interpenetrated shapes.

Usually it makes sense to use a void for the mask.

nb/documentation/interactions_between_geometry/interactions_between_geometry.nb
```diff
@@ -82,6 +97,15 @@ Performs the hull operation on the input shapes.`;
...
+`;
+
+const shaft = Arc(10, 10, 40).color('grey');
+shaft.mask(shaft.grow(1)).view();
```
pentacular, line 107: How about using the simpler form shaft.mask(grow(1)).view() ?

nb/documentation/interactions_with_geometry/interactions_with_geometry.nb
```diff
@@ -38,6 +59,18 @@ Arc(4)
...
+
+md`
+---
+### Fuse
+
+Shapes which are grouped, assembled, or otherwise associated can be fused together to form a single solid
+
+`;
+
+Group(Arc(10), Arc(10).x(6)).fuse().view();
```
pentacular, line 71: We could simplify this as Arc(10).x(0, 6).fuse().view()

nb/documentation/interactions_with_geometry/interactions_with_geometry.nb
```diff
@@ -58,19 +91,28 @@ An item has an exterior (which looks like an individual piece of geometry), and
...
 const aBox = Box(10, 10, 10).as('box');
 
 const aDesign = aBox.fit(aBolt.z(10)).as('design').view();
 
-aDesign.get('*').pack().view().md(`We need to get into the design in order to get at 'box' and 'bolt'`);
+aDesign
+  .in()
+  .pack()
+  .view()
+  .md(`We need to get into the design in order to get at 'box' and 'bolt'`);
```
pentacular, line 112: Should we write "'design' item" to make it clear that it's the item we're getting into?

nb/documentation/reference_geometry/reference_geometry.nb
```diff
@@ -0,0 +1,42 @@
...
+
+Group(a, b, b.to(a.align('<x'))).view();
+
+md`
+---
+### At
+Performes an operation on a shape at a particular location.
+`;
+
+Group(b, a.at(b, cut(b))).view(); //Moves a to b, makes the cut, then moves a back to it's original location
```
pentacular, line 24: How about b.and(a.at(b, cut(b))).view()

nb/documentation/reference_geometry/reference_geometry.nb
```diff
@@ -0,0 +1,42 @@
...
+
+//Group(a, b.at(a, clip(a)).view();
+
+md`
+---
+### By
+Moves a shape from it's current location by a reference amount.
+`;
+
+Group(a, b.by(a)).view(); //Moves the blue square in the X direction the same amount that the red square had been moved
```
pentacular, line 34: How about a.and(b.by(a)).view()?

nb/documentation/reference_geometry/reference_geometry.nb
```diff
@@ -0,0 +1,42 @@
...
+
+md`
+---
+### To
+Moves a shape to a reference position.
+`;
+
+Group(a, b.to(a)).view(); //Moves the blue square to the red square
```
pentacular, line 42: How about a.and(b.to(a)) ?

pentacular:
> 

Very nice -- don't worry about the notebooks too much -- they can be sorted out later. :)

nb/documentation/interactions_between_geometry/interactions_between_geometry.nb
BarbourSmith: Absolutely! I haven't quite gotten used to that yet, but I really should. It's an excellent addition.

nb/documentation/interactions_with_geometry/interactions_with_geometry.nb
BarbourSmith: I like that

nb/documentation/interactions_with_geometry/interactions_with_geometry.nb
BarbourSmith: Yeah, that is more clear

nb/documentation/interactions_between_geometry/interactions_between_geometry.nb
BarbourSmith: I'm not sure I understand the difference. In both cases we want to have something cut a larger hole than it's own size, right?

nb/documentation/interactions_with_geometry/interactions_with_geometry.nb
BarbourSmith: What does Link() do? Maybe I need to document that!

nb/documentation/interactions_between_geometry/interactions_between_geometry.nb
pentacular: Sure, I guess the point is that the relationship between mask and void is accidental rather than fundamental.

BarbourSmith: I see, that is important. It's just taking advantage of the .void feature rather than being an alternative to void.

PR #1130, opened 2022-05-20
Restructure geometry

pentacular:


algorithm/cgal/cgal.cc
```diff
@@ -325,19 +373,43 @@ class SurfaceMeshQuery {
...
+    mesh_.reset(new Surface_mesh(mesh));
+    assert(mesh_->is_valid());
+    tree_.reset(new Tree(faces(*mesh_).first, faces(*mesh_).second, *mesh_));
+    inside_tester_.reset(new Inside_tester(*tree_));
+  }
+
   SurfaceMeshQuery(const Surface_mesh* mesh)
       : is_volume_(CGAL::is_closed(*mesh)) {
     mesh_.reset(new Surface_mesh(*mesh));
+    assert(mesh_->is_valid());
```
BarbourSmith, line 412: What does it mean to check is_valid()?

algorithm/cgal/cgal.cc
```diff
@@ -3460,19 +3586,9 @@ class Geometry {
...
-    std::vector<Polygon_with_holes_2> polygonsWithHoles;
-    gps(nth).polygons_with_holes(std::back_inserter(polygonsWithHoles));
-    for (auto& polygon : polygonsWithHoles) {
-      polygon = transformPolygonWithHoles(polygon, absolute_plane, local_plane,
-                                          absolute_to_local_transform);
-    }
-    emitPlane(local_plane, emit_plane);
-    ::emitPolygonsWithHoles(polygonsWithHoles, emit_polygon, emit_point);
-    plane(nth) = local_plane;
+    assert(is_local_frame());
```
BarbourSmith, line 3589: This is a neat thing to have. Just as a way to catch anything that might slip through in an unexpected frame.

algorithm/cgal/cgal.cc
```diff
@@ -4280,65 +4478,97 @@ int Disjoint(Geometry* geometry, emscripten::val getIsMasked) {
...
           }
-          in.swap(out);
-          out.clear();
         }
-        geometry->segments(start).swap(in);
         break;
       }
       case GEOMETRY_POINTS: {
-        // TBD
+        // TODO: Support disjunction by volumes, segments, polygons.
```
BarbourSmith, line 4571: This is the ability of a say 3D object to cut a point cloud for example?

algorithm/cgal/cgal.cc
```diff
@@ -4974,47 +5206,48 @@ int Join(Geometry* geometry, int targets) {
...
+        for (const Segment& segment : segments) {
+          out.push_back(segment);
         }
         has_last = true;
-        last = input_segments.end()[-1].target();
+        last = segments.back().target();
         break;
       }
       case GEOMETRY_POINTS: {
         // A point is equivalent to a zero-length segment.
```
BarbourSmith, line 5238: Is that really true? I know it's true mathematically, but is the representation of a point done by a zero length segment?

api/shape/Curve.js
```diff
@@ -0,0 +1,13 @@
+import Link from './Link.js';
+import Point from './Point.js';
+import Shape from './Shape.js';
+import bezier from 'adaptive-bezier-curve';
+
+export const Curve = (start, c1, c2, end) => {
```
BarbourSmith, line 6: 👍 👍 👍 👍

api/shape/absolute.js
```diff
@@ -0,0 +1,9 @@
+import Shape from './Shape.js';
+import { makeAbsolute } from '@jsxcad/geometry';
+
+export const absolute = () => (shape) =>
```
BarbourSmith, line 4: So this is a new method a.absolute()?

BarbourSmith (approved):
Looks good to me! I probably won't get to play with it until Sunday, but I am excited to 🙂

algorithm/cgal/cgal.cc
pentacular: Yeah. Pretty much everything should be able to cut points.

algorithm/cgal/cgal.cc
pentacular: Ah, a segment is represented as a pair of points -- source() and target().
A point in this case is a boring x, y, z.

Here we track the last point so that we can form a segment from it to the next point added.

api/shape/absolute.js
pentacular: Yeah. I only exposed it for testing.

It makes the absolute frame of each object into its local frame.

This might be useful, but I'm not sure exactly what for.

BarbourSmith: Sounds good. My next question was going to be "What is it used for?" 🙂

algorithm/cgal/cgal.cc
BarbourSmith: Gotcha, that's what I thought. I was just curious if there was some advantage to using the same representation for segments and for points.

PR #1132, opened 2022-05-24
Add curve to documentation

BarbourSmith:
The second example isn't the simplest, but I thought it was so cool that it belonged in the documentation somewhere.


pentacular:
Very nice. :)

PR #1137, opened 2022-06-03
Clean up cgal code

pentacular:


BarbourSmith (approved):
Sorry it took me so long to read through this, lots of wedding stuff to wrap up. It only took 10 seconds once I actually looked though so I probably could have done it :). Looks good to me!

PR #1138, opened 2022-06-07
Fix segment/volume clip/cut and update cgal

pentacular:


algorithm/cgal/wrap.js
```diff
@@ -0,0 +1,22 @@
+import { STATUS_OK, STATUS_ZERO_THICKNESS } from './status.js';
+import { fromCgalGeometry, withCgalGeometry } from './cgalGeometry.js';
+
+import { ErrorZeroThickness } from './error.js';
+
+export const wrap = (inputs, alpha, offset) =>
```
BarbourSmith, line 6: What is wrap doing? Wrapping a set of points?

api/shape/Hull.js
```diff
@@ -11,7 +11,7 @@ Shape.prototype.Hull = Shape.shapeMethod(Hull);
 export const hull =
   (...shapes) =>
   (shape) =>
-    Hull(shape, ...shapes.map((other) => Shape.toShape(other, shape)));
+    Hull(...shapes.map((other) => Shape.toShape(other, shape)));
```
BarbourSmith, line 14: :)

BarbourSmith (approved):
Looks good to me.

This one looks rotated, not sure if that's correct:

image

algorithm/cgal/wrap.js
pentacular: It wraps a volume around a set of points and/or faces, but it doesn't work very well yet.

PR #1140, opened 2022-06-10
Switch to shared_ptr, add ghost geometry, and approximated vertices i…

pentacular:
…n serialization


es6/jsxcad-ui-threejs.js
```diff
@@ -1183,7 +1192,7 @@ const buildMeshes = async ({
           mesh.layers.set(layer);
           updateUserData(geometry, scene, mesh.userData);
           mesh.userData.tangible = true;
-          if (tags.includes('type:void')) {
+          if (tags.includes('type:ghost')) {
```
BarbourSmith, line 1195: Did ghost completely replace void or is it in addition to void()?

BarbourSmith (approved):
I'm excited to play around with how the ghost geometry feels in practice

PR #1141, opened 2022-06-12
cut and clip leave ghosts; static view defaults to no-grid

pentacular:


es6/jsxcad-ui-threejs.js
```diff
@@ -619,7 +619,13 @@ const merge = async (properties, parameters) => {
 };
 
 const setMaterial = async (definitions, tags, parameters) => {
-  const threejsMaterial = toThreejsMaterialFromTags(tags, definitions);
+  let threejsMaterial;
+  if (tags.includes('type:ghost')) {
+    // FIX: This is ugly.
```
BarbourSmith, line 624: How so?

es6/jsxcad-ui-threejs.js
```diff
@@ -1681,7 +1689,7 @@ const staticDisplay = async (
     canvas,
     geometry,
     withAxes = false,
-    withGrid = true,
+    withGrid = false,
```
BarbourSmith, line 1692: Is the grid gone by default now?

BarbourSmith (commented):
Aren't you supposed to be offline this weekend? 😀 

I like how the ghost looks and I'm excited to see how it feels in practice.

There are a couple of these which are a little concerning

image

Well this is cool:

image

But this doesn't look right, does it?

image

Or this:

image

es6/jsxcad-ui-threejs.js
pentacular: Well, it's just hacking in an arbitrary material if it's a ghost geometry. Actually, a material should probably be set on creation (and transparency set likewise)

es6/jsxcad-ui-threejs.js
pentacular: Yeah, I wanted to see how it looks, and if it helps reduce noise in the diffs.

But an explicitly provided value should be honored.

BarbourSmith: I think that makes sense. On the static views I rarely use the grid, and if I wanted to use it I would probably want the top down view

BarbourSmith (approved):
Is this OK?

image

Same here, does these look right?

image

image

If those look good to you then it all looks good to me!

PR #1142, opened 2022-06-22
Update fromPolygons and ldraw

pentacular:


BarbourSmith (approved):
This generally looks good to me, but I don't fully understand it. What is the advantage of ldraw over fromPolygons? Ldraw is approximate so it's faster?

pentacular:
> This generally looks good to me, but I don't fully understand it. What is the advantage of ldraw over fromPolygons? Ldraw is approximate so it's faster?

Hmm, ldraw uses fromPolygons to produce geometry.

The main changes are making fromPolygons a bit more robust and flexible.

BarbourSmith:
Gotcha, nice :)

PR #1143, opened 2022-06-23
Update static display and fit

pentacular:


es6/jsxcad-ui-threejs.js
```diff
@@ -1375,38 +1382,48 @@ const moveToFit = ({
...
+    new Vector3(box.min.x, box.max.y, box.min.z),
+    new Vector3(box.max.x, box.max.y, box.min.z),
+    new Vector3(box.min.x, box.min.y, box.max.z),
+    new Vector3(box.max.x, box.min.y, box.max.z),
+    new Vector3(box.min.x, box.max.y, box.max.z),
+    new Vector3(box.max.x, box.max.y, box.max.z),
+  ];
 
-  camera.position.copy(center).sub(direction);
+  // Back the camera off by 1mm at a time until the bounding box is contained by
```
BarbourSmith, line 1398: That's neat!

es6/jsxcad-ui-threejs.js
```diff
@@ -1375,38 +1382,48 @@ const moveToFit = ({
...
+  for (let distance = 1; ; distance += 1) {
+    const direction = target
+      .clone()
+      .sub(camera.position)
+      .normalize()
+      .multiplyScalar(distance);
+
+    camera.position.copy(target).sub(direction);
+
+    camera.near = distance / 100;
```
BarbourSmith, line 1410: I was seeing an issue where if I created an orbitView() with a small object in it and then updated it with a larger object the near and far planes would be wrong so part of the object could get cut off. Does this run just once when the orbitView is created, or every time it is updated?

BarbourSmith (approved):
Some of these seem less in frame than before. It seems like specifically flat shapes maybe? 3D stuff seems to look great.

image

es6/jsxcad-ui-threejs.js
pentacular: Just when it's created -- the camera should be persistent through updates, so that you can see changes in-place.

pentacular:
> Some of these seem less in frame than before. It seems like specifically flat shapes maybe? 3D stuff seems to look great.
> 
> image

Ah, maybe it's not contributing to the bounding box properly -- let's see.

pentacular:
> > Some of these seem less in frame than before. It seems like specifically flat shapes maybe? 3D stuff seems to look great.
> > image
> 
> Ah, maybe it's not contributing to the bounding box properly -- let's see.

Good catch -- the world matrix wasn't being updated so the bounding box was taking the unscaled shape.

es6/jsxcad-ui-threejs.js
BarbourSmith: Could we have a way to update just the near and far distances? I'm finding situations where I create a view of something small and then it grows as I work on it. Eventually parts of the geometry stick out past the far plane and aren't rendered anymore

BarbourSmith: That would probably do it. I have made things bigger than 10m (house model).

pentacular: Let's try 1 km, and see how that goes. :)

PR #1144, opened 2022-06-29
Preliminary cut of editing

pentacular:


es6/jsxcad-api-shape.js
```diff
@@ -3318,6 +3318,9 @@ const edit = Shape.chainable(
 
 Shape.registerMethod('edit', edit);
 
+// TODO: Add an option to include a virtual segment at the target of the last
```
BarbourSmith, line 3321: ❤️

es6/jsxcad-api-v1-png.js
```diff
@@ -129,10 +13,7 @@ const readPng = async (path) => {
   return raster;
 };
 
-const readPngAsContours = async (
-  path,
-  { by = 10, tolerance = 5, to = 256 } = {}
-) => {
+const readPngAsContours = async (path, bands = [128, 256]) => {
```
BarbourSmith, line 16: How does this work? It pulls in all the pixels as points?

BarbourSmith (approved):
Looks good to me :)

es6/jsxcad-api-v1-png.js
pentacular: The pixels are converted to contour lines based on the grayscale value -- like on a map.

They come in as segments.

PR #1147, opened 2022-07-16
Integrate remeshing into smooth selections

pentacular:


pentacular:
This also inverts the selection logic for grow, so that the selected regions are grown.

BarbourSmith:
I didn't see the switch to the selection region for grow before, but that makes sense

PR #1148, opened 2022-07-22
Implement loft with constrained polyline triangulation.

pentacular:


BarbourSmith (approved):
Looks good to me. I only saw the changes to loft, I didn't see where the changes to caching happen, where is that fix?

pentacular:
> Looks good to me. I only saw the changes to loft, I didn't see where the changes to caching happen, where is that fix?

Ah, I think that needs a bit more thought now that the issue is better understood.

PR #1149, opened 2022-07-28
Improve eachEdge and orientation

pentacular:


nb/documentation/creating_geometry/creating_geometry.md
```diff
@@ -58,7 +58,7 @@ Curve(Point(0, 0), Point(10, 10), Point(20, 30), Point(40, 0)).view();
 ```JavaScript
 Curve([0, 0], [10, 0], [10, 10], [20, 10])
   .rx(1 / 4)
-  .edges(ArcX(0, 4, 4).hasSides(6), Loft)
+  .eachEdge(ArcX(0, 4, 4).hasSides(6), Loft)
```
BarbourSmith, line 61: I like the name change here. .edges reads nicely, but .eachEdge is more accurate for what it's actually doing :)

BarbourSmith (approved):

nb/documentation/creating_geometry/creating_geometry.md
pentacular: Yeah, it's also in line with points vs eachPoint.

The other thing is that eachEdge is producing a separate geometry for each edge, with its own orientation, as opposed to a set of segments.

PR #1150, opened 2022-08-01
Add profiling output, disjoint strategies, and reduce demeshing

pentacular:


algorithm/cgal/cgal.cc
```diff
@@ -4172,6 +4310,17 @@ int Disjoint(Geometry* geometry, emscripten::val getIsMasked) {
   return STATUS_OK;
 }
 
+int Disjoint(Geometry* geometry, emscripten::val getIsMasked, int mode) {
+  switch (mode == 0) {
```
BarbourSmith, line 4314: What are the differences between these modes?

BarbourSmith (approved):

algorithm/cgal/cgal.cc
pentacular: It's the order in which they're disjointed.

Given A, B, C, backward does A = A.cut(B, C), B = B.cut(C), but forward does B = B.cut(C), A = A.cut(B, C)

On average, backward seems slower, but in some cases it can be significantly faster.

I think it's difficult to intuit performance since it is based on complexity and intersectiveness rather than on volume -- but maybe we can compare the two strategies and get a better feel for what happens.

PR #1161, opened 2022-09-05
Update notebooks and logo

pentacular:


nb/start.md
```diff
@@ -1,4 +1,4 @@
-# JSxCAD
+# Jot (JSxCAD)
```
BarbourSmith, line 1: 😮 Are we fully committing to the name change?

BarbourSmith (approved):
I love the new logo!

nb/start.md
pentacular: Oh, we were committed already -- but I was hoping for a cleaner transition. :)

PR #1163, opened 2022-09-07
Fix controls and refresh issues, make edit/view mode persistent

pentacular:


pentacular:
I've updated alpha with this change to show how it works.

hrgdavor:
image

I do not see how to switch to edit mode from view mode

also showing full url in tab button makes it impossible to have more than few tabs

maybe something like this (for test I just moved path to title attribute):
image

and also showing folder name if there are multiple files with same name

pentacular:
> 



> 



> image
> 
> I do not see how to switch to edit mode from view mode

Try Control-E.

> also showing full url in tab button makes it impossible to have more than few tabs
> 
> maybe something like this (for test I just moved path to title attribute): image
> 
> and also showing folder name if there are multiple files with same name

I agree that it needs improving -- but the solution needs to allow the user to still access the fully qualified path.

On the other hand, I'm not sure that putting the filenames into the tab header is the right solution in the first place. :)

Let's figure out more of what flexlayout-react can do, and see what the cheap options are.

pentacular:
> > 
> 
> > 
> 
> > image
> > I do not see how to switch to edit mode from view mode
> 
> Try Control-E.
> 
> > also showing full url in tab button makes it impossible to have more than few tabs
> > maybe something like this (for test I just moved path to title attribute): image
> > and also showing folder name if there are multiple files with same name
> 
> I agree that it needs improving -- but the solution needs to allow the user to still access the fully qualified path.
> 
> On the other hand, I'm not sure that putting the filenames into the tab header is the right solution in the first place. :)
> 
> Let's figure out more of what flexlayout-react can do, and see what the cheap options are.

Actually, we already have a solution for the fully qualified path -- it is displayed in the navigation bar after the #.

I've published a change to alpha/ so that only the last part of the url is used as the tab title.

ptal

BarbourSmith (approved):

PR #1165, opened 2022-09-08
Update api nb and fix cast/shadow.

pentacular:
This also adjusts opening notebooks.

If the notebook has no edit/view state, it is opened as edit when empty, otherwise as view.


api/shape/shadow.js
```diff
@@ -0,0 +1,19 @@
+import Shape from './Shape.js';
+import { XY } from './refs.js';
+import { cast as castGeometry } from '@jsxcad/geometry';
+
+export const shadow = Shape.chainable(
+  (planeReference = XY(0), sourceReference = XY(1)) =>
+    (shape) =>
+      Shape.fromGeometry(
+        castGeometry(
```
BarbourSmith, line 9: I thought ThreeJS was handling the shadows

BarbourSmith (approved):

api/shape/shadow.js
pentacular: Decorative shadows, yes. :)

These are geometric projections.

I've been a bit unhappy with cast() so I changed it to shadow(), although maybe it should be project().

Let's see how it works in practice.

PR #1167, opened 2022-09-12
More API nb

pentacular:
Adds O


api/shape/get.js
```diff
@@ -27,10 +29,12 @@ export const get = Shape.chainable((...tags) => (shape) => {
   };
   const geometry = shape.toGeometry();
   visit(geometry, walk);
-  return Group(...picks);
+  return groupOp(...picks);
```
BarbourSmith, line 32: What is groupOp?

nb/api/noOp.md
```diff
@@ -0,0 +1,13 @@
+### noOp()
```
BarbourSmith, line 1: No complaining, just wondering. Why do we need this?

BarbourSmith: Oh, I think I get it. It can be passed in place of a function which does something when passing an operation is required?

nb/api/normal.nb
```diff
@@ -0,0 +1,21 @@
+md`
+### normal()
+Produces a reference point along the direction normal to the shape.
+
+This can be visualized as a point.
+
+For complex shapes the normal may be somewhat arbitrary and unintuitive.
```
BarbourSmith, line 7: 😁 What is the normal of a sphere?

nb/api/on.md
```diff
@@ -0,0 +1,39 @@
+### on()
```
BarbourSmith, line 1: I'm not sure I knew about this and it's fantastic!

BarbourSmith (approved):
👍 👍 👍 👍

api/shape/get.js
BarbourSmith: Nevermind...I see where it's defined. I thought it might be a global thing

pentacular: It's a function for grouping the results.

Usually it is Group, but I needed to be able to produce a list sometimes.

get('foo', List)

nb/api/noOp.md
pentacular: Yeah.

nb/api/normal.nb
pentacular: For an approximate sphere it would be approximately arbitrary, I guess. :)

nb/api/on.md
pentacular: Glad to hear it. :)

PR #1169, opened 2022-09-16
More api

pentacular:


api/shape/Plan.js
```diff
@@ -211,6 +211,52 @@ export const getScale = (geometry) => {
   ];
 };
 
+const X = 0;
+const Y = 1;
+const Z = 2;
+
+export const buildCorners = (x, y, z) => {
```
BarbourSmith, line 218: This is a neat idea!

nb/api/Line.md
```diff
@@ -41,11 +41,13 @@ Line(5)
 Line(5).and(rz(1 / 8)) produces a 1/8th angle.
 
 ```JavaScript
-Line(seq({ from: -1, upto: 1, by: 1 / 8 }, noOp, List))
+Line(seq({ from: -1, upto: 1, by: 1 / 8 }, noOp(), List))
```
BarbourSmith, line 44: Why the change to have noOp called instead of passed as a function? I think that the old way makes more sense to me intuitively

nb/api/Spiral.md
```diff
@@ -0,0 +1,54 @@
+### Spiral()
```
BarbourSmith, line 1: Can we do spirals with a z-component at the moment? Just curious

nb/api/SurfaceMesh.md
```diff
@@ -0,0 +1,39 @@
+### SurfaceMesh()
+Parameter|Default|Type
+---|---|---
+serialization||A serialized surface mesh.
+
+Reconstructs a surface mesh from its serialization.
+
+_This is intended for debugging._
```
BarbourSmith, line 8: 👍

nb/api/simplify.md
```diff
@@ -0,0 +1,27 @@
+### simplify()
+Parameter|Default|Type
+---|---|---
+eps||Number of mm for point resolution.
+{ratio}|1|Stop when the remaining edges reaches this fraction.
+
+_Check: does this actually do anything?_
```
BarbourSmith, line 7: 😛

nb/api/size.md
```diff
@@ -0,0 +1,103 @@
+### size()
+Parameter|Default|Type
+---|---|---
+op|size => shape => size|Operates on the size.
+
+Analyzes the shape and calls op with an object with:
```
BarbourSmith, line 6: Does .boundingBox() still exist? I kind of liked having the information returned like thisThingsSize = a.size();

nb/api/smooth.md
```diff
@@ -0,0 +1,22 @@
...
+...selections||Only smooth where covered by a selection shape.
+{iterations}|1|
+{time}|1|
+{remeshIterations}|1|
+{remeshRelaxationSteps}|1|
+
+Smooths shape.
+
+```JavaScript
+Box(10, 10, 10)
```
BarbourSmith, line 14: This is such a cool example

nb/api/sort.md
```diff
@@ -0,0 +1,44 @@
+### sort()
+Parameter|Default|Type
+---|---|---
+spec|'z<y<x<'|String defining the sort predicate.
+
+This orders a group of shapes by their extremal bounding box corner.
```
BarbourSmith, line 6: I didn't know about this one. That's awesome

nb/api/tint.md
```diff
@@ -0,0 +1,34 @@
+### tint()
+Parameter|Default|Type
+---|---|---
+color||The name of the color to tint with.
+
+Sets the color of all leaf geometry in shape to the named color, replacing existing colors.
+
+Mixes the named color with all leaf geometry, merging with existing colors.
```
BarbourSmith, line 8: This seems contradictory to me. Is it replacing the existing color or mixing with the existing color. It's mixing with it, right?

BarbourSmith (approved):
Loving all these descriptions. I'm learning things reading them so I think they will be super helpful to other people.

nb/api/size.md
pentacular: const foo = a.size(noOp()) should get you that, I think.

a.bb() exists and probably should have a long name like a.boundingBox()

nb/api/tint.md
pentacular: Oops, yes.

nb/api/Line.md
pentacular: noOp is a => b => b. So noOp() gets us the b => b which does what we want to pass through the numbers to gather with List.

But maybe this is a bit confusing -- might need to rethink it.

api/shape/Plan.js
pentacular: Yeah, should have factored that out earlier. :)

nb/api/Line.md
BarbourSmith: Ah I understand now. It's a little confusing, but I don't think it's a big deal

nb/api/size.md
BarbourSmith: Nice, that works for me

nb/api/tint.md
pentacular: Yes, you're right -- that first sentence needs deleting.

PR #1175, opened 2022-09-24
Start components

pentacular:
Starts adding parameterized components with interactive example code constructors.


nb/components/shaft.md
```diff
@@ -0,0 +1,129 @@
+### Shaft, Flat, and FlatShaft
```
BarbourSmith, line 1: This is a useful library to have

nb/projects/pentacular/28BYJ-48/28BYJ-48.nb
```diff
@@ -1,56 +1,118 @@
...
 
-export const stepperMotor28byj48 = Group(
-  Arc(bodyDiameter)
-    .add(Box(connectorLength, connectorWidth).x(bodyDiameter / -2))
-    .add(Box(connectorBaseHeight, connectorBaseWidth).x(bodyDiameter / -2 + 2))
-    .ez(-bodyHeight)
-    .mask(grow(bodyPlay))
-    .as('body'),
+md`
+### 28BYJ-48
```
BarbourSmith, line 4: How do users find these? Like if someone want's to use a 28BYJ-48 motor in their project, how do they know that it is an option to import one?

BarbourSmith (approved):

nb/projects/pentacular/28BYJ-48/28BYJ-48.nb
pentacular: That I'm not sure -- I've started having the parametric examples generate example code like:

"import ...; Foo(...)"

But I don't have a good idea for discovery.

BarbourSmith: Is it possible to scrape github somehow? Maybe with a tag?

pentacular: Hmm, could be -- how do github tags work?

BarbourSmith: Ah I think what I really meant is "topics" https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/classifying-your-repository-with-topics

Basically it is an easy to way to group together all the repositories on a topic so if we used the 'jot' topic then we could search or browse within that topic

pentacular: Ah, right -- that's a good idea.

Is there any finer granularity available or just at the repository level?

pentacular: > also if the repo has jot topic, it would be wise to define markup for discovery. for example putting something into package.json so scraper can know where stuff is

That's an interesting idea.

Are there any standards on that or do we just add something ad hoc?

hrgdavor: I know of no specific standards for package.json except the obvious that you want a unique key there, but the object value of your property is I think totally up-to you.

PR #1181, opened 2022-10-14
Various fixes

pentacular:


api/core/recordNotes.js
```diff
@@ -1,4 +1,4 @@
-import { Group, Shape, loadGeometry, saveGeometry } from '@jsxcad/api-shape';
+import { Group, Shape, load, save } from '@jsxcad/api-shape';
```
BarbourSmith, line 1: Is the reason for changing these that the only thing loaded or saved is geometry?

nb/components/shaft.md
```diff
@@ -1,20 +1,44 @@
...
 
 ```JavaScript
-const length = control('length', 10, 'input');
+export const Shaft = ({ diameter = 10, length = 10, play = 0.2 } = {}) =>
+  Arc(diameter).ez(length).clean().mask(grow(play)).as('shaft');
 ```
 
 ```JavaScript
-const flatLength = control('flat length', 5, 'input');
+export const FlatShaft = ({
```
BarbourSmith, line 14: Is this the new format for making controls?

nb/projects/pentacular/tiles/hex.md
```diff
@@ -0,0 +1,380 @@
+```JavaScript
+const hex = Hexagon(32);
+```
+
+```JavaScript
+const Tab = (thickness, tabLength) =>
```
BarbourSmith, line 6: I feel like tabs are a common enough thing that I do that it might be worth having a stand alone tabs library that we could import and use for things like this

BarbourSmith: Although it's not that hard to do them as is

BarbourSmith (approved):
Looks great to me!

nb/projects/pentacular/tiles/hex.md
pentacular: Yes, I think so.

I think we're a bit weak on geometric introspection at the moment.

I'll try to get better support for finding where rays intersect volumes, etc, for more flexible placement.

PR #1184, opened 2022-11-21
Asynchronous chaining

pentacular:


api/shape/eachPoint.js
```diff
@@ -4,25 +4,27 @@ import Shape from './Shape.js';
...
+      let [pointOp = (point) => (shape) => point, groupOp = Group] =
+        shapesAndFunctions;
+      const coordinates = [];
+      let nth = 0;
+      eachPointOfGeometry(await shape.toGeometry(), ([x = 0, y = 0, z = 0]) =>
+        coordinates.push([x, y, z])
+      );
+      const points = [];
+      for (const [x, y, z] of coordinates) {
+        points.push(await pointOp(Point().move(x, y, z), nth++))
```
BarbourSmith, line 21: This is interesting, I haven't seen github do this kind of thing before

api/shape/toFlatValues.js
```diff
@@ -0,0 +1,25 @@
+import Shape from './Shape.js';
+import { toValue } from './toValue.js';
+
+export const toFlatValues = Shape.registerMethod(
```
BarbourSmith, line 4: What does toFlatValues do?

BarbourSmith: I guess the opposite of toNestedValues, but when do we want to use these?

nb/projects/BarbourSmith/belt_sprockets/belt_sprockets.md
```diff
@@ -7,9 +7,8 @@ const additionalToothDepth = 0; //mm
 ```
 
 ```JavaScript
-const ToothSpaceingCurvefit = (teeth, b, c, d) => {
-  return ((c * Math.pow(teeth, d)) / (b + Math.pow(teeth, d))) * teeth;
-};
+const ToothSpaceingCurvefit = (teeth, b, c, d) =>
```
BarbourSmith, line 10: This is just a style change right? No functional difference here?

nb/projects/BarbourSmith/belt_sprockets/belt_sprockets.nb
```diff
@@ -27,15 +27,14 @@ const additionalToothDepth = 0; //mm
...
-};
+const ToothSpaceingCurvefit = (teeth, b, c, d) =>
+  ((c * Math.pow(teeth, d)) / (b + Math.pow(teeth, d))) * teeth;
 
 const ToothSpacing = (teeth, tooth_pitch, pitch_line_offset) =>
   2 * ((teeth * tooth_pitch) / (3.14159265 * 2) - pitch_line_offset);
 
 const ToothProfileToFace = (...points) =>
-  Shape.fromOpenPath(points).loop().fill();
+  Shape.chain(Shape.fromOpenPath(points)).loop().fill();
```
BarbourSmith, line 37: Chain is nice here

nb/projects/BarbourSmith/belt_sprockets/belt_sprockets.nb
```diff
@@ -289,9 +276,7 @@ export const T10 = {
     [2.971997, 0.007239, 0],
     [3.06511, 0, 0],
     [3.06511, -1, 0]
-  )
-    .md('T10 profile')
```
BarbourSmith, line 293: Why did these go away?

BarbourSmith (approved):
I think I'll have to use it in practice to really have any thoughts, but I'm excited to try it :)

api/shape/Arc.js
```diff
@@ -100,65 +95,67 @@ const reifyArc =
...
-
-Shape.prototype.Arc = Shape.shapeMethod(Arc);
-Shape.prototype.ArcX = Shape.shapeMethod(ArcX);
-Shape.prototype.ArcY = Shape.shapeMethod(ArcY);
-Shape.prototype.ArcZ = Shape.shapeMethod(ArcZ);
+const reifyArcZ = reifyArc(Z);
+const reifyArcX = reifyArc(X);
+const reifyArcY = reifyArc(Y);
+
+const ArcOp =
```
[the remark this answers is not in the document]
pentacular: Ah, it's mostly because I haven't cleaned up the constructor / reifier split properly from before.

Those should be able to be merged back together again.

api/shape/Hexagon.js
```diff
@@ -1,8 +1,8 @@
 import { Arc } from './Arc.js';
 import Shape from './Shape.js';
 
-export const Hexagon = (x, y, z) => Arc(x, y, z).hasSides(6);
-
-Shape.prototype.Hexagon = Shape.shapeMethod(Hexagon);
+export const Hexagon = Shape.registerShapeMethod('Hexagon', (x, y, z) =>
```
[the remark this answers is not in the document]
pentacular: Yeah, should have done that earlier. :)

api/shape/Plan.js
```diff
@@ -17,59 +20,87 @@ const scale = (amount, [x = 0, y = 0, z = 0]) => [
...
+  [bx = 0, by = 0, bz = 0]
+) => [ax - bx, ay - by, az - bz];
+
+export const computeMiddle = (c1, c2) => [
+  (c1[X] + c2[X]) * 0.5,
+  (c1[Y] + c2[Y]) * 0.5,
+  (c1[Z] + c2[Z]) * 0.5,
+];
+
+export const computeSides = (c1, c2, sides, zag = 0.01) => {
```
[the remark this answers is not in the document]
pentacular: Yeah, a lot of the plan stuff needs to be ripped out still, but I think we're getting cleaner primitives.

api/shape/eachPoint.js
pentacular: Which kind of thing?

api/shape/toFlatValues.js
pentacular: That's a good question -- I think those need to be cleaned up and rationalized.

nb/projects/BarbourSmith/belt_sprockets/belt_sprockets.md
pentacular: Yeah.

nb/projects/BarbourSmith/belt_sprockets/belt_sprockets.nb
pentacular: This is more of a hack to avoid thinking about how to fix it more cleanly. :)

The problem is that Loop doesn't support receiving raw coordinates, but fromOpenPaths does.

Need to think about how to rationalize all of that as well -- the parameter conversion stuff is a bit too magical.

nb/projects/BarbourSmith/belt_sprockets/belt_sprockets.nb
pentacular: Ah, they're getting invoked indirectly when-ever you build these -- e.g., from examples -- so you'd get 20 copies of each for each example.

Generally you want these things at the example level rather than the reuse level.

nb/projects/BarbourSmith/belt_sprockets/belt_sprockets.nb
BarbourSmith: How does the parameter conversion stuff work as it is?

pentacular: toShape is the most common conversion -- if it receives a function it will apply the current shape to it.
toShapes is similar, but will handle Group, and Array.
toCoordinate tries to produce [x, y, z], and so on.

Those are reasonably straight-forward.

Some others probably need rethinking, like toNestedValues, etc.

api/shape/eachPoint.js
pentacular: Ah, yes -- that is nice.

PR #1185, opened 2022-11-29
Fix some breakage

pentacular:


BarbourSmith (approved):
I'm sorry it took me so long to read this :/ 

It looks like there is something going on with scale for sure, but other than that it looks good to me.

Lots of changes that look like this:

image

pentacular:
> I'm sorry it took me so long to read this :/
> 
> It looks like there is something going on with scale for sure, but other than that it looks good to me.
> 
> Lots of changes that look like this:
> 
> image

Yeah, orbs were twice as big as they ought to have been. :)

PR #1186, opened 2022-12-10
Rename mask to masked, use LoadXxx with path based constructors, make fonts work as urls.

pentacular:


convert/stl/fromStl.js
```diff
@@ -13,10 +13,7 @@ const toParser = (format) => {
   }
 };
 
-export const fromStl = async (
-  stl,
-  { format = 'ascii', geometry = 'graph' } = {}
-) => {
+export const fromStl = async (stl, { format = 'ascii' } = {}) => {
```
BarbourSmith, line 16: What if it's binary?

nb/api/Spiral.nb
```diff
@@ -0,0 +1,37 @@
+md`
+[index](#https://raw.githubusercontent.com/jsxcad/JSxCAD/master/nb/api/index.nb)
+### Spiral()
+Parameter|Default|Type
+---|---|---
+op|Point|Function or shape to build from.
```
BarbourSmith, line 6: The documentation on these is really nice :)

BarbourSmith (approved):
After a refresh I'm still not seeing some images, but what I can see looks great 👍 👍

convert/stl/fromStl.js
pentacular: This is the default, binary can be passed in.
Probably it should check that it's one of the accepted formats.

nb/api/Spiral.nb
pentacular: Yeah, moving it to the table form helped a lot, since they're mostly not positional arguments now.

PR #1187, opened 2022-12-17
Add unfold()

pentacular:
This adds unfold() and diameter(), cleans up some output paths, and removes api/v1-pdf, v1-svg, v1-stl


hrgdavor:
where can unfold be seen in action ? I am just curious :)

algorithm/cgal/Cast.h
```diff
@@ -0,0 +1,43 @@
+int Cast(Geometry* geometry) {
```
BarbourSmith, line 1: Liking these new multiple files :)

BarbourSmith (approved):
Breaking the C++ stuff into multiple files was such a good idea. It's way easier to understand now I think

BarbourSmith:
@hrgdavor It's still very much a work in progress I believe but the basic example is something like this from the PR:

image

Basically taking a 3D shape and "unfolding" into a 2D version, ideally the most compact 2D version with the minimum number of joints.

pentacular:
> @hrgdavor It's still very much a work in progress I believe but the basic example is something like this from the PR:
> 
> image
> 
> Basically taking a 3D shape and "unfolding" into a 2D version, ideally the most compact 2D version with the minimum number of joints.

Yes, it produces a set of faces and fold edges, which can then be post-processed.

This is the kind of thing you could produce from those faces and edges for e.g., printing.

image

PR #1188, opened 2023-01-03
Move some constructors to general operators and clean up toolpaths

pentacular:


algorithm/cgal/Geometry.h
```diff
@@ -0,0 +1,662 @@
...
+  Points& input_points(int nth) {
+    if (!has_input_points(nth)) {
+      input_points_[nth].reset(new Points);
+    }
+    return *input_points_[nth];
+  }
+
+  bool has_points(int nth) { return points_[nth] != nullptr; }
+
+  std::vector<Point>& points(int nth) {
```
BarbourSmith, line 234: Is it normal to have this stuff in the the .h file? Not complaining, just trying to learn. I think of .h files as being definitions only

api/shape/gcode.js
```diff
@@ -0,0 +1,38 @@
...
+  (...args) =>
+    async (shape) => {
+      const {
+        value: name,
+        func: op = (s) => s,
+        object: options = {},
+      } = destructure(args);
+      const { id, path, viewId } = qualifyViewId(name, getSourceLocation());
+      let index = 0;
+      // Gcode doesn't do layout (although maybe it should)
```
BarbourSmith, line 20: 😛

convert/gcode/toGcode.js
```diff
@@ -1,15 +1,18 @@
...
+) => {
   const codes = [];
 
   // CHECK: Perhaps this should be a more direct modeling of the GRBL state?
   const state = {
     // Where is the tool
-    x: undefined,
-    y: undefined,
-    z: undefined,
+    x: 0,
```
BarbourSmith, line 13: I think that this is a reasonable assumption

convert/gcode/toGcode.js
```diff
@@ -1,15 +1,18 @@
...
+    y: 0,
+    z: 0,
     // How 'fast' the tool is running (rpm or power).
     s: 0,
     f: 0,
```
BarbourSmith, line 18: These defaulting to zero is probably the safest bet, but I have seen situations where if F defaults to zero when you try to run a program that doesn't set F the machine locks up and never moves...which is probably the safest result, but it's confusing since it seems like the machine is broken.

BarbourSmith (approved):
Looks great to me. Sometimes it's a little bit hard to tell what's new and what's just moved from one file to a different one, but it is nice to read it all because even the parts that I'm pretty sure aren't new it's nice to understand better.

algorithm/cgal/Geometry.h
pentacular: Traditionally .h files just contain the interfaces, but with header libraries the definitions need to be there too.

api/shape/gcode.js
pentacular: Ah, we can include layout now that it's segment based.

convert/gcode/toGcode.js
pentacular: Hmm, interesting.

I've cleaned this up a bit to only set to a changed F on demand of a cut operation.

Let's see if that will work better.

PR #1190, opened 2023-01-06
Convert registerShapeMethod cases to registerMethod

pentacular:
This allows constructors to take context.

e.g., shape.Box(diameter()) will construct a box with the diameter of shape.


api/shape/Arc.js
```diff
@@ -165,9 +171,9 @@ const ArcOp =
     return result;
   };
 
-export const Arc = Shape.registerShapeMethod('Arc', ArcOp('Arc'));
-export const ArcX = Shape.registerShapeMethod('ArcX', ArcOp('ArcX'));
-export const ArcY = Shape.registerShapeMethod('ArcY', ArcOp('ArcY'));
-export const ArcZ = Shape.registerShapeMethod('ArcZ', ArcOp('ArcZ'));
+export const Arc = Shape.registerMethod('Arc', ArcOp('Arc'));
```
BarbourSmith, line 174: Nice. Shape.registerShapeMethod seems redundant 😛

geometry/tagged/computeToolpath.js
```diff
@@ -51,10 +51,12 @@ export const computeToolpath = (
     stopCost = 30,
     candidateLimit = 1,
     subCandidateLimit = 1,
+    radialCutDepth = toolDiameter / 4,
```
BarbourSmith, line 54: 👍 This is a good default...1/3 would probably be safe too

BarbourSmith (approved):
👍 👍 

There are a couple visual diff changes that I'm not 100% sure about...not that they are wrong, just that I wasn't sure. Like this sort of thing:

image

pentacular:
> +1 +1
> 
> There are a couple visual diff changes that I'm not 100% sure about...not that they are wrong, just that I wasn't sure. Like this sort of thing:
> 
> image

Yeah, I think this one is correct now -- it should produce a group with both parts of the cut.

Not sure how it got to the earlier state -- need to be a bit more careful. :)

api/shape/Arc.js
pentacular: Yeah, it was necessary before the new chaining system was in effect, but now we can remove it as a special case. :)

PR #1191, opened 2023-01-13
Re-enable mesh approximation, improved soup repair, oriented bounding boxes, fixes bend

pentacular:


algorithm/cgal/fromPolygonSoup.js
```diff
@@ -0,0 +1,30 @@
+import { STATUS_OK, STATUS_ZERO_THICKNESS } from './status.js';
+import { fromCgalGeometry, withCgalGeometry } from './cgalGeometry.js';
+
+import { ErrorZeroThickness } from './error.js';
```
BarbourSmith, line 4: I kinda haven't run into any zero thickness issues in a while, how did you fix that again?

BarbourSmith (approved):
Looks great to me :)

algorithm/cgal/fromPolygonSoup.js
pentacular: cgal still rejects zero thickness, but manifold permits it.

The booleans are defaulting to manifold at the moment, so it's probably silently passing them along, and they're just not manifesting as problems.

I see some zero thickness parts in my outputs, but they're relatively benign.

PR #1192, opened 2023-01-24
Add initial occt support (disabled)

pentacular:


algorithm/cgal/Clip.h
```diff
@@ -1,3 +1,5 @@
+#include "BRepAlgoAPI_Common.hxx"
```
BarbourSmith, line 1: What is a .hxx file?

algorithm/cgal/EagerTransform.h
```diff
@@ -11,8 +11,24 @@ int EagerTransform(Geometry* geometry, int count) {
...
     switch (geometry->getType(nth)) {
       case GEOMETRY_MESH: {
-        CGAL::Polygon_mesh_processing::transform(
-            transform, geometry->mesh(nth), CGAL::parameters::all_default());
+        // Ideally we would:
+        //  auto r = transformGeometryOfOcctShape(transform, geometry->occt_shape(nth));
+        //  geometry->setOcctShape(nth, r);
+        //  geometry->discard_mesh(nth);
+        // but transformGeometryOfOcctShape is not working reliably.
+        // So for now we drop down to the cgal mesh representation where we need a general transform
```
BarbourSmith, line 19: 👍

algorithm/cgal/Geometry.h
```diff
@@ -128,17 +142,54 @@ class Geometry {
...
+    return meshes_[nth] != nullptr;
+  }
 
   Surface_mesh& mesh(int nth) {
     if (!has_mesh(nth)) {
       meshes_[nth].reset(new Surface_mesh);
+#ifdef ENABLE_OCCT
+      if (has_occt_shape(nth)) {
+        // TODO: Pick a better default tolerance.
+        buildSurfaceMeshFromOcctShape(occt_shape(nth), /*tolerance=*/0.01, *meshes_[nth]);
```
BarbourSmith, line 166: What does the tolerance do here? Also I really like this style of comment, I didn't know you could do that

algorithm/cgal/occt_util.h
```diff
@@ -0,0 +1,188 @@
...
+  return solid_maker.Solid();
+}
+
+void buildSurfaceMeshFromOcctShape(const TopoDS_Shape& shape, double deflection_tolerance, Surface_mesh& mesh) {
+  Vertex_map vertices;
+  for(TopExp_Explorer it(shape, TopAbs_SOLID) ; it.More() ; it.Next()){
+    TopoDS_Solid solid = TopoDS::Solid(it.Current());
+    BRepMesh_IncrementalMesh(solid, deflection_tolerance);
+    // Each face is triangulated separately.
+    // We could try to consolidate vertices, but let's see if we get t-junctions first.
```
BarbourSmith, line 83: How did this play out?

BarbourSmith (approved):
Looks good to me :)

algorithm/cgal/occt_util.h
pentacular: Ah, that needs updating -- we don't see t junctions because the triangulation is entirely controlled by geometric shape, unlike in jscad.

algorithm/cgal/Clip.h
pentacular: An old convention for c++ headers. :)

algorithm/cgal/Geometry.h
pentacular: According to https://dev.opencascade.org/doc/refman/html/class_b_rep_mesh___incremental_mesh.html#a3a383b3afe164161a3aa59a492180ac6 it is the linear deflection.

So how far out of true points may be from the ideal curve.

PR #1193, opened 2023-01-28
Clean up and archive occt support

pentacular:
Unable to isolate the buffer overflow issues from occt -- may be due to interaction with wasm-exceptions.

Plan to revisit in a few months.


BarbourSmith (approved):
Sounds like a good plan to me :)

What happens now if it is enabled? Does everything break, or it only breaks if we actually try to use the occt functions?

pentacular:
> Sounds like a good plan to me :)
> 
> What happens now if it is enabled? Does everything break, or it only breaks if we actually try to use the occt functions?

It only breaks if you try to use some occt functions -- you can create occt spheres and boxes, for example.

Subtracting occt spheres or boxes from those will fail, but substracting non-occt spheres and boxes from occt spheres and boxes will succeed, etc.

PR #1196, opened 2023-02-11
Add shell and dilateXY

pentacular:


algorithm/cgal/Shell.h
```diff
@@ -0,0 +1,243 @@
...
+    double edge_size = 1;
+
+    Mesh_criteria criteria(p::facet_angle = angle, p::facet_size = sizing,
+                           p::facet_distance = approx,
+                           p::facet_topology = topology,
+                           p::edge_size = edge_size);
+
+    bool maybe_hollow = inner_offset > -rad2;
+
+    if (protect) {
```
BarbourSmith, line 159: 👍

api/shape/cut.js
```diff
@@ -1,13 +1,38 @@
...
+    args,
+    'geometry',
+    'geometries',
+    'modes'
+  );
+  return Shape.fromGeometry(
+    cutGeometry(
+      first,
+      rest,
+      modes.includes('open'),
```
BarbourSmith, line 17: What do these do?...Or rather when would the user use them?

BarbourSmith (approved):
Does this change seem right?

image

If that is OK it looks great to me

pentacular:
> Does this change seem right?
> 
> image
> 
> If that is OK it looks great to me

I temporarily disabled toolpaths :( So it's expected for now.

api/shape/cut.js
pentacular: The different between a.cut and a.Cut is if a is included implicitly in the cut, or just used as context.

a.cut(b, c) === a.Cut(self(), b, c)

One example where you might use a.Cut is a.Cut(offset(1), inset(1))

pentacular: This should be similarly true across the rest of the operations -- e.g., and, And, etc.

BarbourSmith: That makes a lot of sense...nice!

PR #1197, opened 2023-02-22
Simple non-overlapping toolpathing

pentacular:


algorithm/cgal/ComputeToolpath.h
```diff
@@ -28,201 +34,265 @@ int ComputeToolpath(Geometry* geometry, size_t material, size_t selection,
...
-              mesh, vertices,
-              Point((x + 1.0) * tool_spacing, (y - 0.0) * tool_spacing, height)),
-          ensureVertex(
-              mesh, vertices,
-              Point((x + 1.0) * tool_spacing, (y + 1.0) * tool_spacing, height)),
-          ensureVertex(
-              mesh, vertices,
-              Point((x - 0.0) * tool_spacing, (y + 1.0) * tool_spacing, height)));
-    }
+    auto PickJumpCell = [&](Cell* current_cell, double& next_cost,
```
BarbourSmith, line 209: This all looks good, but why is it in the .h file? I thought that .h files were only supposed to contain definitions? (Not complaining, just trying to learn)

BarbourSmith (approved):
👍 👍

algorithm/cgal/ComputeToolpath.h
pentacular: Just the usual header library approach.

One problem is that if we put this into a .cc file we may find that a lot of cgal gets duplicated between the various .o files coming out.

But it does make it slower to recompile.

BarbourSmith: Ah right, I've asked this before. I remember now :)

PR #1201, opened 2023-05-16
Separate align and alignment, and add hole as an alias for void.

pentacular:


api/shape/alignment.js
```diff
@@ -0,0 +1,94 @@
+import Point from './Point.js';
```
BarbourSmith, line 1: I like the name change for this file

api/shape/void.js
```diff
@@ -3,7 +3,9 @@ import { hasTypeGhost, hasTypeVoid } from '@jsxcad/geometry';
 import Shape from './Shape.js';
 
 export const voidFn = Shape.registerMethod(
-  'void',
+  ['void', 'hole'],
```
BarbourSmith, line 6: Hole is nice

nb/api/align.md
```diff
@@ -25,40 +25,40 @@ Box(10, 10, 1).view().note('Box(10, 10, 1)');
...
+Box(10, 10, 1).align('x>').and(origin).view().md("align('x>')");
 ```
 
 ![Image](align.md.$5.png)
 
 align('x<')
 
 ```JavaScript
-Box(10, 10, 1).by(align('x<')).and(origin).view().md("align('x<')");
+Box(10, 10, 1).align('x<').and(origin).view().md("align('x<')");
```
BarbourSmith, line 47: This is way easier to read I think

BarbourSmith (approved):
Looks good to me!

api/shape/void.js
pentacular: Yeah, the critical problem is the inability to write void() :)

nb/api/align.md
pentacular: Yes, I think so -- using alignment() should be an unusual case.

PR #1202, opened 2023-05-23
Various fixes

pentacular:


api/core/evaluate.js
```diff
@@ -128,7 +128,8 @@ export const execute = async (
                 completed.add(id);
                 console.log(`Completed ${id}`);
               } catch (error) {
-                throw error;
+                // This should be reported via a note.
```
BarbourSmith, line 131: What are notes?

es6/jsxcad-ui-app.js
```diff
@@ -2860,6 +2860,36 @@ class EditNote extends ReactDOM$3.Component {
   }
 }
 
+class ErrorNote extends ReactDOM$3.Component {
```
BarbourSmith, line 2863: What do these look like in practice?

BarbourSmith (approved):
Does this look OK?

image

Other than that looks great to me 👍 👍

api/core/evaluate.js
pentacular: Ah, notes are the elements that notebooks are made of. :)

pentacular:
> Does this look OK?
> 
> image
> Other than that looks great to me +1 +1

Thanks for catching that. :)

I forgot to uncomment something.

es6/jsxcad-ui-app.js
pentacular: Red text like "TypeError: Box(...).ez(...).clean(...).And(...).clip(...).flub is not a function"

PR #1203, opened 2023-05-25
Add fetch timeouts, optimize fill, add simple math, refactor size, fix some uses of align.

pentacular:


BarbourSmith (approved):
You are on a roll!

Does the way that these two shapes are filling look right? Other than that everything looks good to me :)

image

image

pentacular:
> You are on a roll!
> 
> Does the way that these two shapes are filling look right? Other than that everything looks good to me :)
> 
> image image

Yeah, this has been bothering me for a while -- I need to dig into the underlying problem -- but it is pre-existing.

PR #1204, opened 2023-06-09
Initial migration to registerMethod2

pentacular:
Starts work on separating out the flexible and simple api interfaces.


api/core/evaluate.js
```diff
@@ -25,11 +25,7 @@ export const evaluate = async (ecmascript, { api, path }) => {
     );
     // Add import to make import.meta.url available.
     const op = await builder({ ...api, import: { meta: { url: path } } });
-    // Retry until none of the operations block.
-    // Note: This retry mechanism should be obsolete now.
```
BarbourSmith, line 29: 👍

api/shape/Plan.js
```diff
@@ -46,12 +46,12 @@ export const computeSides = (c1, c2, sides, zag = 0.01) => {
 export const zagSides = Shape.registerMethod2(
   'zagSides',
   ['number', 'number'],
-  (diameter = 1, zag = 0.01) => 1 / toSidesFromZag(diameter, zag)
+  (diameter = 1, zag = 0.01) => toSidesFromZag(diameter, zag)
```
BarbourSmith, line 49: What does this change mean? Why was it 1/ before...ah was the number of segments?

api/shape/destructure.js
```diff
@@ -333,7 +379,35 @@ export const destructure2 = async (shape, input, ...specs) => {
...
+          } else if (Shape.isArray(value) && value.every(Shape.isSegment)) {
+            out.push(...value);
+          } else {
+            rest.push(arg);
+          }
+        }
+        output.push(out);
+        break;
+      }
+      case 'coordinateLists': {
```
BarbourSmith, line 406: Has "coordinates" fully replaced "points"?

api/shape/nth.js
```diff
@@ -1,35 +1,33 @@
...
-        if (nth < 0) {
-          nth = candidates.length - nth;
-        }
-        const candidate = candidates[nth];
-        if (candidate === undefined) {
-          group.push(Empty());
-        } else {
-          group.push(candidate);
-        }
+  ['input', 'numbers'],
```
BarbourSmith, line 8: How do these work? Is this for documentation?

BarbourSmith (approved):
Wow, this is a big pull request :)

Looks great to me!

api/shape/Plan.js
pentacular: I managed to confuse it with zagStep :)

api/shape/nth.js
pentacular: This tells it how to do the magical decoding stuff.

This means, "look for an input shape first, then look for numbers".

But, yeah, it should also be used to generate documentation and UI automatically.

These then get passed in with the corresponding arguments to the actual function (which should move out of here and into geometry)

api/shape/destructure.js
pentacular: I've started using coordinates for things like [1, 2, 3] as opposed to Point()

PR #1205, opened 2023-06-12
Clean up view and convert more methods

pentacular:


api/shape/sort.js
```diff
@@ -7,65 +7,65 @@ const X = 0;
...
+export const sort = Shape.registerMethod2(
   'sort',
-  (spec = 'z<y<x<') =>
-    async (shape) => {
-      let leafs = [];
-      for (const leaf of getLeafs(await shape.toGeometry())) {
-        const [min, max] = measureBoundingBox(leaf);
-        const shape = await Shape.fromGeometry(leaf);
-        leafs.push({ min, max, shape });
+  ['inputGeometry', 'string'],
```
BarbourSmith, line 12: Are you converting all of these by hand or is there some sort of macro to do it?

BarbourSmith (approved):
I'm not sure if these missing diffs are indicative of some sort of issue (probably not, probably just github), but other than that looks great to me 👍 👍 

image

api/shape/sort.js
pentacular: Manually, for my sins. :)

Hopefully it will get simpler over time.

BarbourSmith: 😬 Good luck!

PR #1206, opened 2023-06-13
Upgrade more methods and fix view issue

pentacular:
viewId was incorrectly receiving options in some cases


BarbourSmith (approved):
👍 👍 Looks good to me!

algorithm/cgal/fromPolygonSoup.js
```diff
@@ -24,6 +24,13 @@ export const fromPolygonSoup = (
       cornerThreshold,
     })}`
   );
+  if (cornerThreshold === 0) {
+    try {
```
BarbourSmith, line 28: What does this do?

algorithm/cgal/FromPolygonSoup.h
```diff
@@ -22,32 +28,32 @@ int FromPolygonSoup(Geometry* geometry, emscripten::val fill) {
...
-        std::vector<Face_index> faces;
-        CGAL::Polygon_mesh_processing::triangulate_hole(
-            mesh, edge, std::back_inserter(faces),
-            CGAL::parameters::use_2d_constrained_delaunay_triangulation(false));
-        if (faces.empty()) {
-          failed = true;
-        }
-        break;
-      }
+  if (wrap_always || CGAL::Polygon_mesh_processing::does_self_intersect(mesh)) {
```
BarbourSmith, line 31: How fast is wrapping?

algorithm/cgal/computeOrientedBoundingBox.js
```diff
@@ -6,25 +6,29 @@ export const computeOrientedBoundingBox = (inputs) => {
...
-        );
-      case STATUS_EMPTY:
-        return;
-      default:
-        throw new Error(`Unexpected status ${status}`);
+  return withCgalGeometry(
+    'computeOrientedBoundingBox',
+    inputs,
+    (cgalGeometry, g) => {
+      const status = g.ComputeOrientedBoundingBox(cgalGeometry);
```
BarbourSmith, line 13: My brain read this as "Computer oriented bounding box" the first time through and I was like "What is the computer frame of reference? Doesn't it move with the camera? 🙃

api/shape/shadow.js
```diff
@@ -4,9 +4,15 @@ import { cast as castGeometry } from '@jsxcad/geometry';
 
 export const shadow = Shape.registerMethod2(
   'shadow',
```
BarbourSmith, line 6: Did this used to be called project? I feel like the name shadow might imply some sort of parallax where there is a point light source...although maybe I'm over thinking it and 'project' has the same issue

BarbourSmith (approved):
Looks great to me!

api/shape/shadow.js
pentacular: I think it probably ought to be silhouette(). wdyt?

algorithm/cgal/FromPolygonSoup.h
pentacular: On my machine, it's tens of seconds for most shapes, which isn't fast, but is usable.

algorithm/cgal/fromPolygonSoup.js
pentacular: Skips simplification if all corners should be preserved.

api/shape/shadow.js
BarbourSmith: Brilliant. That's exactly the right word

algorithm/cgal/fromPolygonSoup.js
BarbourSmith: but why does it throw an error and then catch it?

pentacular: Ah, it gives you a stack trace to figure out how you got there. :)

I was trying to figure out what wasn't propagating the threshold through properly.

PR #1207, opened 2023-06-20
Complete registerMethod2 migration

pentacular:


algorithm/cgal/FromPolygonSoup.h
```diff
@@ -5,55 +5,75 @@ int FromPolygonSoup(Geometry* geometry, emscripten::val fill, bool wrap_always,
...
+      std::cout << "QQ/FromPolygonSoup/Wrap: alpha=" << alpha
+                << " offset=" << offset << std::endl;
+      std::cout << "QQ/FromPolygonSoup/Wrap/MeshClear" << std::endl;
+      mesh.clear();
+      std::cout << "QQ/FromPolygonSoup/Wrap/Wrap" << std::endl;
+      wrap_epick(points, faces, alpha, offset, mesh);
+    }
 
-  demesh(mesh);
+    if (corner_threshold > 0) {
```
BarbourSmith, line 66: This is a nice way to do it. No need for a simplification flag and a corner threshold value

ui/app/App.js
```diff
@@ -497,6 +497,18 @@ class App extends React.Component {
     };
 
     this.Notebook.run = async (path, options) => {
+      if (this.Notebook.runDebounce) {
```
BarbourSmith, line 500: What are we debouncing? I think of that as being a hardware thing when a button is noisy

BarbourSmith (approved):
Congratulations on getting it all done :)

ui/app/App.js
pentacular: Same problem, I think -- something is triggering two simultaneous runs, but I haven't managed to track it down yet.

PR #1208, opened 2023-06-21
Re-enable using chains as functions.

pentacular:


BarbourSmith (approved):
Looks good to me. Its nice to see the awaits gone

PR #1209, opened 2023-06-24
Initial migration of api logic to geometry

pentacular:


api/shape/Arc.js
```diff
@@ -97,7 +98,7 @@ const reifyArc =
       }
     }
 
-    return spiral.absolute();
+    return absolute()(spiral);
```
BarbourSmith, line 101: How does this syntax work?

api/shape/Arc.js
```diff
@@ -160,26 +161,27 @@ const ArcOp =
...
     const [c1, c2] = await buildCorners(x, y, z)(null);
-    const result = reify({ c1, c2, start, end, sides, zag });
-    return result;
+    const result = await reify({ c1, c2, start, end, sides, zag });
+    const geometry = await result.toGeometry();
+    return geometry;
   };
 
-export const Arc = Shape.registerMethod2(
+export const Arc = Shape.registerMethod3(
```
BarbourSmith, line 169: Three! 😃

api/shape/Curve.js
```diff
@@ -7,15 +7,19 @@ import Shape from './Shape.js';
...
+  ) => {
     const interpolator = new CurveInterpolator(coordinates, {
-      closed: isClosed,
+      closed,
       tension: 0.2,
       alpha: 0.5,
     });
     const points = interpolator.getPoints(steps);
-    if (isClosed) {
+    if (closed) {
```
BarbourSmith, line 22: I like this name change 👍

api/shape/asPart.js
```diff
@@ -1,12 +1,11 @@
 import Shape from './Shape.js';
-import { taggedItem } from '@jsxcad/geometry';
+import { asPart as op } from '@jsxcad/geometry';
```
BarbourSmith, line 2: Why the switch to all of these being 'op' now? Just more condensed?

api/shape/view.js
```diff
@@ -14,22 +14,22 @@ const MODES =
   'modes:grid,none,side,top,wireframe,noWireframe,skin,noSkin,outline,noOutline';
 
 const applyModes = async (shape, options, modes) => {
-  if (modes.includes('wireframe')) {
+  if (modes.wireframe) {
```
BarbourSmith, line 17: This is nicer and cleaner 😍

nb/api/Icosahedron.md
```diff
@@ -8,7 +8,7 @@ Produces an icosahedron that fits the bounding dimensions.
 Icosahedron(3)
 
 ```JavaScript
-Icosahedron(3).and(Box(3, 3, 3).material('glass')).view().md('Icosahedron(3)');
+Icosahedron(3).and(Box(3, 3, 3).material('glass')).view().note('Icosahedron(3)');
```
BarbourSmith, line 11: 👍

api/shape/Arc.js
pentacular: Yeah, once everything is up to 3, I think it'll be time to remove the numbers. :)

api/shape/Arc.js
pentacular: At the api/shape level, operators are functors.

f = op(args) builds a function which waits for an input shape for context.
r = f(input) then calls it.

This allows us to pass around things like cut(1/4) and then later choose what shape to apply it to.

Here we're just saying, call absolute with no parameters on spiral.

PR #1216, opened 2023-09-08
Updates CGAL, breaks loft, makes notebooks unordered

pentacular:
Unfortunately Loft needs a new implementation, which will be a bit big to include.

Notebooks are now unordered which allows code like:
const a = Foo();
const Foo = () => Box();
This allows more freedom in notebook organization.


BarbourSmith (approved):
What does it mean that the notebook is unordered? Is it just populating the top level variables in any order, but then if I have:

let a = 1
log(a)
a = 2
log(a) 

It's still going to print:

1
2

Right?

I finally read MaddAddam, it was great and much more optimistic than I expected

pentacular:
> What does it mean that the notebook is unordered? Is it just populating the top level variables in any order, but then if I have:
> 
> let a = 1 log(a) a = 2 log(a)
> 
> It's still going to print:
> 
> 1 2

Well, given that we make top level definitions immutable, no. :)

But when evaluating those top level definitions, we include their dependencies in topological order.

So it should still produce the same results regardless of the ordering.

e.g.,

const a = Foo();
const Foo = Box;

is reorganized as
const Foo = Box;
const a = Foo();
when evaluating a.

> 
> Right?
> 
> I finally read MaddAddam, it was great and much more optimistic than I expected

Yes, I thought so, too. :)

BarbourSmith:
Tricky! 

How does it map out all the dependencies to reorder?

pentacular:
> Tricky!
> 
> How does it map out all the dependencies to reorder?

We run an AST analysis to identify the non-local variables in use.

This is also used to identify the minimum set of code required to evaluate an individual top-level definition.

BarbourSmith:
AST analysis?

pentacular:
> AST analysis?

Abstract Syntax Tree.

e.g., a parsed representation of the code which shows which bits are variables, functions, expressions, etc.

Like this: https://astexplorer.net/

PR #1217, opened 2023-09-11
Arrange notebooks as sorted accordions

pentacular:


BarbourSmith (approved):
Surprisingly concise for such a big change!

pentacular:
> 

Yes, it's nice for a change. :)

But it's also because it's a bit of a quick hack -- have to see how annoying it is before investing in a more principled solution.

PR #1219, opened 2023-09-18
UI

pentacular:
Update notebook ui (also changes default view size).

Update simplify algorithm to consider interim angles.

Move large stls to jsxcad.js.org


es6/jsxcad-api-shape.js
```diff
@@ -1297,6 +1297,19 @@ const clipFrom = Shape.registerMethod3(
   clipFrom$1
 );
 
+const cloud = Shape.registerMethod3(
```
BarbourSmith, line 1300: Is cloud a point cloud?

BarbourSmith (approved):
This one looked big when I saw the number of files changed, but that was just because of the change in default image size 😂

es6/jsxcad-api-shape.js
pentacular: Yes, just a set of points sampled at a given resolution.

PR #1222, opened 2023-10-26
Sectionized notebooks

pentacular:
Removes source round-trip and makes each section a set of results -- code, views, controls, downloads, etc -- as a discrete card.

Updates for a card are unitary rather than incremental.

Table of contents now indicates which sections are being processed.

Refs are changed from points to items.

.to now moves an object such that a reference point on the source is moved to the target reference point.


api/core/evaluate.js
```diff
@@ -9,7 +9,7 @@ import {
 import { getApi } from './api.js';
 import { toEcmascript } from '@jsxcad/compiler';
 
-export const evaluate = async (ecmascript, { api, path }) => {
+export const evaluate = async (ecmascript, { api, id, path }) => {
```
BarbourSmith, line 12: What is the ID used for?

nb/test/test.md
```diff
@@ -1,3 +1,94 @@
+![Image](test.md.joint.png)
```
BarbourSmith, line 1: This looks cool! What is it?

BarbourSmith (approved):
Looks good to me 👍 👍

api/core/evaluate.js
pentacular: Ah, it tells us which definition we're evaluating.
e.g., for const foo = bar(); id === 'foo'.

nb/test/test.md
pentacular: I've been making a stand for the steam deck, and wanted a joint that could fold into a triangle.

Although I realized a little while ago that I modeled it incorrectly -- instead of a joint with two arms, I should be modeling a half-hinge with an arm. :)

api/core/evaluate.js
BarbourSmith: Ahhh that makes sense

PR #1223, opened 2023-10-28
Support get into items and fix hold semantics

pentacular:
Also make unproductive gets fail by default.


api/shape/sort.js
```diff
@@ -46,10 +42,10 @@ export const sort = Shape.registerMethod3(
       if (limit !== undefined) {
         switch (order) {
           case '>':
-            list = list.filter(({ min }) => min[axis] > limit);
+            leafs = leafs.filter(({ min }) => min[axis] > limit);
             break;
           case '<':
-            list = list.filter(({ max }) => max[axis] < limit);
+            leafs = leafs.filter(({ max }) => max[axis] < limit);
```
BarbourSmith, line 48: Nice name improvement!

api/shape/view.js
```diff
@@ -53,7 +53,14 @@ export const baseViewOp = async (
   geometry,
   name,
   op = (_x) => (s) => s,
-  { size = 256, inline, width, height, position = [100, -100, 100] } = {}
```
BarbourSmith, line 56: That's cool that there is an option to download views

doc/updateNotebook.js
```diff
@@ -69,47 +69,52 @@ const writeMarkdown = async (
...
+                  `cp '${observedPath}' '${expectedPath}'`
+                );
+              }
+            } catch (error) {
+              console.log(`EE/3: ${JSON.stringify(error)}`);
+              if (error.code === 'ENOENT') {
+                failedExpectations.push(
+                  `cp '${observedPath}' '${expectedPath}'`
+                );
+                failedExpectations.push(`git add '${expectedPath}'`);
```
BarbourSmith, line 100: What is the notebook github integration like?

geometry/hold.js
```diff
@@ -1,6 +1,13 @@
 import { Group } from './Group.js';
-import { inItem } from './inItem.js';
-import { on } from './on.js';
 
-export const hold = (geometry, geometries) =>
-  on(geometry, inItem(geometry), (inside) => Group([inside, ...geometries]));
+export const hold = (geometry, geometries) => {
```
BarbourSmith, line 3: What does hold do?

BarbourSmith (approved):
Just questions, not really any suggestions or anything, but it looks great to me!

doc/updateNotebook.js
pentacular: Mostly trying to delegate that to the desktop tools.

Adding local directory access next.

geometry/hold.js
pentacular: Ah, it causes an item to include another piece of content.

PR #1224, opened 2023-11-02
Initial local file access

pentacular:


BarbourSmith (approved):
Awesome! From a user perspective does supporting local files mean that users can edit files on their computer directly through the web interface?

pentacular:
> Awesome! From a user perspective does supporting local files mean that users can edit files on their computer directly through the web interface?

Not yet. :)

This just allows you to read http://a/b/c off e.g., ~/c

But, yes, it should be extended to allow direct editing of source and automatic output of downloads, etc.

hrgdavor:
what is used to run the standalone app ?

pentacular:
> what is used to run the standalone app ?

See https://jsxcad.js.org/nb/index.html for an example.

PR #1227, opened 2023-11-29
Add 'split' mode for minimizeOverhang and improve local directory access

pentacular:


geometry/convexHull.js
```diff
@@ -18,13 +18,16 @@ export const ConvexHull = (geometries) => {
 export const convexHull = (geometry, geometries) =>
   ConvexHull([geometry, ...geometries]);
 
-export const ChainConvexHull = (geometries) => {
+export const ChainConvexHull = (geometries, { close = false } = {}) => {
```
BarbourSmith, line 21: Close links the last one to the first one in the chain?

ui/app/App.js
```diff
@@ -605,7 +605,7 @@ class App extends React.Component {
 
     this.Notebook.run = async (path, options) => {
       const { LocalFilesystemHandles = [] } = this.state;
-      for (const { handle } of LocalFilesystemHandles) {
+      for (const [, handle] of LocalFilesystemHandles) {
```
BarbourSmith, line 608: What does the , here do?

BarbourSmith (approved):
Very slick! 👍 👍

geometry/convexHull.js
pentacular: Yeah, the same works with Link() which makes Loop() redundant -- I should deprecate Loop.

ui/app/App.js
pentacular: ([a, b] = [1, 2]) assigns 1 to a and 2 to b.

([, b] = [1, 2]) ignores 1 and assigns 2 to b.

PR #1229, opened 2023-12-14
Use ace editor again and save sections for reuse

pentacular:


BarbourSmith (approved):
Looks good to me. What are the advantages of the Ace editor?

pentacular:
> Looks good to me. What are the advantages of the Ace editor?

Ah, it's just going back to the original editor, but with the widget support removed.

The idea of just using editable elements was interesting, but didn't work so well in practice. :)

BarbourSmith:
Ah gotcha, it was a cool experiment :)

PR #1230, opened 2023-12-21
Add route and skeleton.

pentacular:
Route provides a vertically oriented solid-along-segment sweep operation, supporting concave shapes.

Skeleton generates a set of skeleton segments for a volume.


nb/api/skeleton.md
```diff
@@ -0,0 +1,21 @@
+[index](../../nb/api/index.md)
+### skeleton()
+Parameter|Default|Type
+---|---|---
+
+Generates a set of skeletal segments from the volumes of the input geometry.
```
BarbourSmith, line 6: What would these be used for?

BarbourSmith (approved):
Route looks particularly cool :)

nb/api/skeleton.md
pentacular: There are a couple of possibilities.

* Finding excessively thin parts of a model and impose a minimum diameter.
* Producing armatures for articulation.
* Segmentation of models into main parts.

Not sure how well it will work in practice yet, though.

BarbourSmith: Very cool ideas :)

PR #1231, opened 2023-12-27
Updates approximate and smooth, and adds refine and fair, extends validate.

pentacular:
Note that refine is problematic due to non-coplanar edge-flips.


algorithm/cgal/fair.js
```diff
@@ -3,9 +3,23 @@ import { fromCgalGeometry, withCgalGeometry } from './cgalGeometry.js';
 
 import { ErrorZeroThickness } from './error.js';
 
-export const fair = (inputs, count, density = 0) =>
+export const fair = (
```
BarbourSmith, line 6: To what extent are the default parameters OK? I worry a little bit about folks getting started being intimidated by tuning things

nb/api/iron.md
```diff
@@ -0,0 +1,18 @@
+[index](../../nb/api/index.md)
+### iron()
```
BarbourSmith, line 2: This is my favorite new feature in a while

BarbourSmith (approved):
Looks awesome to me! 👍 👍 


This one is a super neat example: 

image

algorithm/cgal/fair.js
pentacular: Yeah, that'll take a little experimentation to figure out -- but hopefully there should be only one principle knob.

PR #1233, opened 2024-01-18
Pass options through to Hexagon, etc, add Masked, support inline sequence generation, remove grid, swap { to } and { upto / downto } in seq.

pentacular:
A bit cluttered.

The biggest change is to allow things like a.rz({ by: 1/3 }) instead of a.rz(0/3, 1/3, 2/3).

Along with this, switched the default sequence generation to exclude the endpoint.
{ by: 1/3, upto: 1 } would produce 0/3, 1/3, 2/3, 3/3.


BarbourSmith:
That is very elegant!

nb/projects/pentacular/lego_sorter/lego_sorter.md
```diff
@@ -3,8 +3,8 @@
 ```JavaScript
 const grid = Hexagon(16)
   .seq(
-    { from: -4, to: 4 },
-    { from: -4, to: 4 },
+    { from: -4, upto: 4 },
```
BarbourSmith, line 6: This is more clear 👍 👍

api/shape/maskedBy.js
```diff
@@ -0,0 +1,14 @@
+import Shape from './Shape.js';
+import { maskedBy as op } from '@jsxcad/geometry';
+
+export const maskedBy = Shape.registerMethod3(
```
BarbourSmith, line 4: How does this work?

BarbourSmith (approved):
Looks great to me!

api/shape/maskedBy.js
pentacular: A mask subtracts its own shape instead of the shape it masks.

So you can, e.g., have a drill bit masked by a cylinder -- if you join that, you'll get the drill. If you subtract that, you'll remove a cylindrical cavity.

Mostly useful for adding play around an object.

nb/projects/pentacular/lego_sorter/lego_sorter.md
pentacular: Yeah, it was annoying me for a long time. :)

api/shape/maskedBy.js
BarbourSmith: Ah that's right. I remember this now :)

PR #1234, opened 2024-02-14
Resync

pentacular:
The main impact is to remove the visualization grid, and provide better support for png loading.

Unfortunately bunch of stuff has been accidentally mixed together here.


BarbourSmith:
What does this one do?

BarbourSmith (approved):
Does this change how the UI looks or just how it's working underneath? 

Those textures are looking really good.

PR #1235, opened 2024-02-23
Move api/v1-dxf into api/shape

pentacular:


api/shape/dxf.js
```diff
@@ -0,0 +1,43 @@
+import { fromDxf, toDxf } from '@jsxcad/convert-dxf';
+import { getSourceLocation, read, write } from '@jsxcad/sys';
+import { gridView, qualifyViewId } from './view.js';
+
+import { Shape } from './Shape.js';
+import { ensurePages } from '@jsxcad/geometry';
+
+export const LoadDxf = Shape.registerMethod3(
```
BarbourSmith, line 8: Does this add support for loading and exporting DXF files?

BarbourSmith (approved):

api/shape/dxf.js
pentacular: Mostly it moves it into the core and updates the interface a bit.

Not all of DXF is supported, though -- needs support for SPLINE, at least.

BarbourSmith: DXF seems like a remarkably fragile format given how widely supported it is. I'm had tons of issues with different version of DXF not playing well together

PR #1236, opened 2024-02-27
Update gauge to scale numbers and allow configurable needle length.

pentacular:


BarbourSmith (approved):
What do gauges do?

pentacular:
> What do gauges do?

They display the diameter of the shape in some dimension.

e.g., showing that a box is 54 mm long.

It's not a great name, unfortunately.

BarbourSmith:
I'm 99% sure I asked the same exact thing before. For some reason gauge in my mind is a dial, but I also can't think of a better word.

PR #1238, opened 2024-03-04
Remove callbacks from Geometry and simplify exact representations

pentacular:


BarbourSmith (approved):
This looks nice:

image

This looks maybe wrong:

image

Same with a couple of these actually...

What is it supposed to do?

image

pentacular:
Yes, these are wrong. :)

I put it up as "wip" because I was having trouble seeing what was failing.

Those diffs should be gone now.

BarbourSmith:
Very nice :) 

Looks great to me 👍 👍

BarbourSmith (approved):

algorithm/cgal/Unfold.h
```diff
@@ -73,15 +79,15 @@ static int Unfold(
                   computeInverseSegmentTransform(s1, s2, Vector(0, 0, 1));
               geometry->segments(edge).emplace_back(s1, s2);
               geometry->setTransform(edge, t.inverse());
-              emit_tag(edge, std::string("unfold:edge"));
+              tags.emplace_back(edge, "unfold:edge");
```
BarbourSmith, line 82: What is the current status of unfold?

BarbourSmith (approved):

algorithm/cgal/Unfold.h
pentacular: Works, but haven't really had much use of it so far.

PR #1239, opened 2024-03-15
Compile and run native code via node-addon-api bindings and use to run regression tests in CI

pentacular:
This has revealed problems with loft which will need to be fixed in follow-up.


BarbourSmith (approved):
It looks like a lot of work catching all those little differences 👍 👍

PR #1241, opened 2024-03-21
Re-enable manifold

pentacular:


BarbourSmith (approved):
Looks good to me 👍 👍

hrgdavor:
will there be a jsxcad npm package, and I guess a wasm module inside it ? main readme links to some outdated stuff like google groups and a google doc for API.

pentacular:
> will there be a jsxcad npm package, and I guess a wasm module inside it ? main readme links to some outdated stuff like google groups and a google doc for API.

Ah, good point -- I will clean up that readme.

As for npm packaging, I think it isn't a priority at the moment.

And we keep meaning to rename this. :)

PR #1242, opened 2024-04-15
Initial CNC integration

pentacular:


ui/app/App.js
```diff
@@ -1298,6 +1334,29 @@ class App extends React.Component {
...
+      const tabNode = tabsetNode.getChildren()[selected];
+      if (tabNode.id === 'Make/Tabset/Tab') {
+        return;
+      }
+      const { ip, type } = tabNode.getConfig();
+      const cnc = getCnc({ ip, type });
+      if (path && !data) {
+        data = await readOrWatch(path, { workspace });
+      }
+      await cnc.uploadAndRun(data);
```
BarbourSmith, line 1357: 🥇

BarbourSmith (approved):
This looks awesome!

Are you at all worried about turning off CORS from a security perspective?

ui/app/App.js
pentacular: I don't think there are any security implications.

The CNC side hasn't changed, and the browser is sending blind, and listening for feedback from the websocket.

It just makes the conversations a bit more awkward.

BarbourSmith: Couldn't some random website take control of your CNC and turn it on?

PR #1243, opened 2024-04-22
Add simple websocket based server and restrict compiler to explicit api.

pentacular:


server/ws/server.js
```diff
@@ -0,0 +1,22 @@
+import { WebSocketServer } from 'ws';
+import { Worker } from 'worker_threads';
+
+var server = new WebSocketServer({
+  port: 8080,
+});
+
+const start = async () => {
+  let nextSessionId = 0;
+  server.on('connection', async (ws) => {
```
BarbourSmith, line 10: If the connection drops is there any way to pick back up?

BarbourSmith (approved):
Super cool!

server/ws/server.js
pentacular: Not at the moment -- I used the connection drop to reset the session.

But it could be done with a TTL.

PR #1244, opened 2024-04-28
Remove cgal.h

pentacular:
Breaks up the headers into mostly independent units.


BarbourSmith (approved):
Looks good to me 👍 👍 

The advantage of this is to keep things organized?

pentacular:
> Looks good to me 👍 👍
> 
> The advantage of this is to keep things organized?

Mostly -- it's still a bit of a mess, but the dependencies are clearer now.

Also it was just too big. :)

PR #1245, opened 2024-04-30
Support point and segment extrusion

pentacular:


nb/api/bend.nb
```diff
@@ -13,7 +13,7 @@ See: [remesh](#https://raw.githubusercontent.com/jsxcad/JSxCAD/master/nb/api/rem
 `;
 
 Box(40, 1, 5)
-  .and(bend(20))
+  .And(y(20), bend(20))
```
BarbourSmith, line 16: Does capital And do anything different?

algorithm/cgal/Bend.h
```diff
@@ -3,49 +3,128 @@
 #include "demesh_util.h"
 #include "point_util.h"
 
-static int Bend(Geometry* geometry, double reference_radius) {
+static int Bend(Geometry* geometry, double reference_radius,
```
BarbourSmith, line 6: What does bend do again?

BarbourSmith (approved):
Looking at bend made me think again about how to flatten out bent wood shapes and I did some research on how other CAD programs do it and it seems like for the most part they are doing a conversion to mesh and then flattening the mesh

nb/api/bend.nb
pentacular: The capital forms only use the input shape for context.

So this produces two versions of Box(40, 1, 5) -- one translated, and one bent.

The lowercase form would produce three versions -- the original, the translated, and the bent.

algorithm/cgal/Bend.h
pentacular: Applies a uniform curvature to the shape.

Effectively placing it along the perimeter of a circle of the given radius.

It's not quite what we generally want, though.

PR #1246, opened 2024-05-07
Generalize grow, refactor math, random

pentacular:
Grow/Trim

This mostly changes grow to do a minkowski sum with a convex hull upon points, segments, polygons, and meshes.

Trim is added as a lower dimensional counterpoint -- given a mesh it grows the individual faces, given a polygon it grows the individual edges.

This allows, e.g., x.cut(trim(Orb(1))) to trim x by 1 mm in all dimenions.

The tool can be two dimensional, e.g., x.grow(Arc(1)) to grow in x and y by one mm without affecting z.

The tool can be asymmetric, e.g., x.grow(Box(1)) to grow in x and y by one mm, with sharp axis-aligned edges.

Or x.grow(Orb(cut(XY()))) to grow rounded in all directions other than positive z.

Random
The main use of random numbers so far is to produce a set of N values.

x.random(4, 12, (a, b, c, d) => ...) generates the first four numbers of seed 12 and binds them to a, b, c, d.

The Nth set of M random numbers can be selected with the third integer argument.

e.g., Seq({ to: 20 }, t => random(4, 12, t, (a, b, c, d) => ...))

will repeat 20 times using different groups of 4 random values.


BarbourSmith:
The tool can be asymmetric, e.g., x.grow(Box(1)) to grow in x and y by one mm, with sharp axis-aligned edges.

Or x.grow(Orb(cut(XY()))) to grow rounded in all directions other than positive z

This is slick

algorithm/cgal/Grow.h
```diff
@@ -1,96 +1,226 @@
...
 
 #include "Geometry.h"
+#include "manifold_util.h"
+#include "transform_util.h"
 
-static int Grow(Geometry* geometry, size_t count, bool x, bool y, bool z) {
-  typedef CGAL::Exact_predicates_exact_constructions_kernel EK;
-  typedef CGAL::Surface_mesh<EK::Point_3> Surface_mesh;
-  size_t size = geometry->size();
+static int Grow(Geometry* geometry, size_t count) {
```
BarbourSmith, line 20: What does count do here?

algorithm/cgal/computeBoundingBox.js
```diff
@@ -19,7 +19,7 @@ export const computeBoundingBox = (inputs) => {
       case STATUS_EMPTY:
         return;
       default:
-        throw new Error(`Unexpected status ${status}`);
+        throw new Error(`Unexpected status ${status} in computeBoundingBox`);
```
BarbourSmith, line 22: These are nice 👍

BarbourSmith (approved):
Is it intensional that these bounding boxes all went away?

image

Does this look right?

image

algorithm/cgal/Grow.h
pentacular: Count is the number of geometries in the input (as opposed to the tool).

There's probably a better naming choice for that.

pentacular:
> Is it intensional that these bounding boxes all went away?

Yes, I finally deleted all of the existing ones to force generation, since the bounding boxes were sometimes under the pixel threshold.

I'm hoping that gauges provide a better alternative.

> 
> image
> Does this look right?
> 
> image

Ah, thanks for noticing that.

I've removed fromPolygons and fixed fromPolygonSoup to make sure it bounds a volume when closed.

PR #1249, opened 2024-06-29
Adds an alternative loft implementation.

pentacular:
This loft implementation operates by considering the lofting as a recursive decomposition of a triangle strip, which prevents local self-intersection, and picks the best solution for each strip at each point, avoiding incremental errors.

However, it just makes a guess at which cycles should be connected, and while it requires that all triangle bases be consumed, it does not require that all triangle tips be consumed.

This is a step forward, but more work is required.


BarbourSmith:
What are the issues with the old version of loft that this is addressing?

algorithm/cgal/Loft2.h
```diff
@@ -0,0 +1,321 @@
...
+#include "Geometry.h"
+#include "demesh_util.h"
+
+typedef CGAL::Exact_predicates_exact_constructions_kernel EK;
+
+static bool loft_between_cycles(std::vector<EK::Segment_3>& lower_segments,
+                                std::vector<EK::Segment_3>& upper_segments,
+                                std::vector<EK::Point_3>& points,
+                                std::vector<std::vector<size_t>>& faces) {
+  // Each segment forms a triangle with one point in the other set.
```
BarbourSmith, line 23: Nice succinct description 👍 👍

algorithm/cgal/Loft2.h
```diff
@@ -0,0 +1,321 @@
...
+    CGAL::Polygon_mesh_processing::reverse_face_orientations(positive_mesh);
+    positive_mesh.join(negative_mesh);
+  }
+
+  int target = geometry->add(GEOMETRY_MESH);
+  geometry->setIdentityTransform(target);
+  demesh(positive_mesh);
+  geometry->mesh(target) = std::move(positive_mesh);
+  return STATUS_OK;
+}
```
BarbourSmith, line 321: This is a super cool loft algorithm. Did you find it somewhere or create it?

BarbourSmith (approved):
👍 👍

algorithm/cgal/Loft2.h
pentacular: Unfortunately I had to figure it out, which is why it took a while :)

The basic idea is pretty easy -- you find the best triangle to bridge the gap between the loops.
The hard part is making sure that it doesn't intersect with any existing triangles.

What I eventually realized is that you can think of each triangle formation as a cut operation.
The first cut splits the loops into strips.
After that, each new triangle splits that strip into two sub-strips.

This also allows the best triangles to be found independently, which avoids build-up error.

But there are still some vulnerabilities.

Any number of triangles can meet at the same tip, so if you have inclined surfaces being lofted, all of the bottom surface will want to join at the bottom-most point of the top surface, and so on.

Having them take turns seems to help break up these patterns, but there are still degenerate cases coming up.

BarbourSmith: That is super cool! You should publish that idea somewhere because I think that has some general applicability

pentacular: Might be an idea, but should understand these failure cases better first. :)

PR #1251, opened 2024-10-11
update manifold

pentacular:


algorithm/cgal/install_manifold_native.sh
```diff
@@ -1,6 +1,6 @@
 ([[ -d manifold ]] || git clone https://github.com/pentacular/manifold &&
   cd manifold &&
-  git checkout jot
+  # git checkout jot
```
BarbourSmith, line 3: The name change is trickling in 😀

BarbourSmith (approved):
I'm pretty out of the loop, but it looks good to me! 

A couple of the visual diffs look a little funky:

image

image

image

How are things going generally?

pentacular:
> How are things going generally?

Trying to figure out why a bunch of stuff is broken after upgrading manifold. :)

It may be that I'm still not processing them quite right.

I'm also thinking of working on a cli interface for a bit to support batch processing without browser.

pentacular:
> > How are things going generally?
> 
> Trying to figure out why a bunch of stuff is broken after upgrading manifold. :)
> 
> It may be that I'm still not processing them quite right.
> 
> I'm also thinking of working on a cli interface for a bit to support batch processing without browser.

Ok, this mostly seems to be working now. :)

BarbourSmith:
A CLI would be pretty sweet for doing bulk jobs. That would probably open up the possibility of some new use cases

PR #1253, opened 2024-10-25
Simple raycasted rendering to generate png

pentacular:


api/shape/png.js
```diff
@@ -103,3 +119,90 @@ export const LoadPngAsRelief = Shape.registerMethod3(
     );
   }
 );
+
+export const png = Shape.registerMethod3(
```
BarbourSmith, line 123: Wild! This is awesome.

BarbourSmith (approved):
Just out of curiosity why PNG over SVG?

pentacular:
> Just out of curiosity why PNG over SVG?

I was thinking about how to generate renders without requiring a browser in the loop.
I realized that I could just cast a matrix of rays at the geometry and turn those into pixels.
It wouldn't be as good, but it would probably be sufficient.

I used png because this technique produces a raster of pixels.

svg is a much harder problem -- you could z-sort flattened triangles, but you might end up with a very large number of triangles, or you could compute the envelope which would give the visible triangles only, but is quite expensive.

Or you could compute the contours in the raster and convert those to svg -- but maybe that would undermine the point of svg since it would introduce flaws that would become worse as you scale up.

btw,  the png render is the nb/api/png.icosahedron.png image.
You can see that it doesn't do the edges or color yet. :)

PR #1254, opened 2024-10-27
Clean up booleans and handle consistent fallback from manifold to cgal.

pentacular:


pentacular:
> The fallback is just if something goes wrong with manifold?

For now, but if manifold fails, it should mean that the inputs are not manifold.
Ideally we should go and check what's wrong with them and figure out some strategy.
Perhaps failing and forcing it to be fixed upstream is the right solution.

> 
> Does this one look right?
> 
> image

I thought it did, but now I think I need to take another look -- thanks.

PR #1256, opened 2024-11-22
Initial support for mirroring source to a spreadsheet

pentacular:
This allows a spreadsheet to be used as a kind of distributed filesystem.

Later we should support three-way merges.

As this filesystem is limited to source code, performance and capacity demands are low


BarbourSmith (approved):
This is so cool!!

Do sheets support diffs and merging things together? Why sheets and not something like docs?

pentacular:
> This is so cool!!
> 
> Do sheets support diffs and merging things together? Why sheets and not something like docs?

They don't support merging directly, but they do have history.
The main reason for sheets is that it fits a key-value store model.
So we can set up columns like 'path', 'value', etc.

I think docs would make more sense for publishing from notebooks, but not so much for a filesystem kind of thing.

BarbourSmith:
Totally, that makes sense. 

Awesome!

PR #1257, opened 2024-11-25
Fix vertex representation and adjust camera range

pentacular:
Large numbers were using exponential notation, breaking objects larger than about one meter.

The camera was also focused too close making large structures difficult to see.

Also introduces @preact/signals to make persistent config easier to delegate.


BarbourSmith:
You are on a roll!! 👍 👍

ui/app/SheetStorage.js
```diff
@@ -233,6 +243,32 @@ export class SheetStorage {
   }
 }
 
+class Config {
+  constructor(path, defaultValue = '') {
+    this.defaultValue = defaultValue;
+    this.path = path;
+    this.signal = signal(defaultValue);
```
BarbourSmith, line 250: What do signals do?

ui/app/package.json
```diff
@@ -36,7 +36,8 @@
     "flexlayout-react": "^0.5.20",
     "hash-sum": "^2.0.0",
     "marked": "^4.0.10",
-    "preact": "10.5.14",
+    "preact": "10.25.0",
+    "@preact/signals": "1.3.0",
```
BarbourSmith, line 40: Ah I guess I can answer my own question 😛

BarbourSmith (approved):
It's exciting to see things moving so quick.

Have you been modeling anything cool lately?

ui/app/SheetStorage.js
pentacular: They're variables which trigger re-rendering on change.

They seem like they might be an improvement over passing state all over the place.

I'm hoping that they allow much of the logic to become decentralized.

pentacular:
> It's exciting to see things moving so quick.
> 
> Have you been modeling anything cool lately?

I'm not sure it's cool, but I have been modeling some desk design / room layout.

image

Although that's an old version which doesn't have the room layout in it.

But it did show me problems with large objects. :)

BarbourSmith:
That is looking awesome!

BarbourSmith:
How is packing working these days?

PR #1258, opened 2024-11-27
Remove puppeteer and switch to headless rendering to png

pentacular:


BarbourSmith (approved):
Other than a tiny bit of rasterization these all look fantastic!

BarbourSmith:
Is the rasterization deterministic so it won't make the test fail when comparing the output? It should be, right?

pentacular:
> Other than a tiny bit of rasterization these all look fantastic!

Yeah, I can't quite figure it out, except that perhaps it's related to devicePixelRatio causing the browser to oversample.

pentacular:
> Is the rasterization deterministic so it won't make the test fail when comparing the output? It should be, right?

I hope so, but mostly I wanted to (a) remove the dependency on pupeteer, and (b) support thumbnail generation.

So now we should be able to say shape.png('foo') and be able to get a png of that view of shape.

PR #1259, opened 2024-11-30
Redesign sort() and add desk design.

pentacular:
Sort needed to be more generic.

Strangely I am getting very different packing results in node and browser for the desk example.


BarbourSmith (approved):

BarbourSmith:
Hmmm that is interesting. Both are deterministic though?

pentacular:
> Hmmm that is interesting. Both are deterministic though?

They ought to be -- more than that, there should be significant biases toward having the same result.

I'll have to figure this out in a bit.

PR #1260, opened 2024-12-03
Lazy matrix composition

pentacular:
Matrix composition seemed to be taking up about 25% of used time.

Here we compute the matrix composition on demand, and remove the approximate matrix from the representation.


BarbourSmith (approved):
Nice optimization! How much of a speed difference does it make?

